PR #14, opened 2015-03-05
Schema remove version array

ericwa:
Well, it works! It's not exactly beautiful...

Stuff that deserves attention:
- domainsToMigrateForItem has been overhauled.
- There's a lot in the test suite that calls [self checkObject: migratedChild versionForDomain: @"Test"] etc. These are doing less than before, because if the domain you pass in isn't the same as the COItem's domain, I just look up the version in the metamodel. So it's not really checking that the migration happened. Not sure if we need more checks that the COItem actually came out correct.
- The API for registering "old metamodel shapshots" looks like this:
```
	[COSchemaMigration recordVersionsByDomain: @{ @"Test" : @(0),
												  @"org.etoile-project.CoreObject" : @(0) }
									forDomain: @"Test"
									  version: 0
								   entityName: @"OutlineItem"];
```

The last step which I didn't do, because it's purely an aesthetic change and would make the change harder to review, is rename kCOObjectVersionsProperty and kCOObjectDomainsProperty (and the underlying strings).


SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -42,28 +42,81 @@ static inline void addObjectForKey(NSMutableDictionary *dict, id object, NSStrin
...
 	ETEntityDescription *entity = [_modelDescriptionRepository descriptionForName: item.entityName];
 	BOOL isDeletedEntity = (entity == nil);
 	
 	/* For a deleted entity, the domain versions match between item and packages */
 	if (isDeletedEntity)
 	{
-		/* The first domain is the one owning the entity, the remaining domains
-		   own the parent entities. */
-		return S([domains firstObject]);
+		// FIXME: We hit this line when loading a CODictionary during the test suite,
```
qmathe: It's expected we hit this line for CODictionary. CODictionary doesn't have a corresponding ETEntityDescription. It is an entity name created at serialization time when encoutering a NSDictionary. At deserialization time, when an item is declared as CODictionary, it will be deserialized as a NSDictionary.

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -42,28 +42,81 @@ static inline void addObjectForKey(NSMutableDictionary *dict, id object, NSStrin
...
+		// something must be wrong (but no tests fail).
+		return S(item.packageName);
+	}
+	
+	/* Early exit, the item has no version specified, or no package, (i.e.
```
qmathe: An important point is that we don't support package deletion, see -testPackageRenaming that explains why. 
So the case where the item has no package gives us compatibility with old CoreObject versions, e.g. the current  release, prior to the schema migration support. 

However in the long run, once we have done the next release (and migrated our existing stores), we should surely remove this and requires the package name to be never nil. We could add a TODO note.

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -42,28 +42,81 @@ static inline void addObjectForKey(NSMutableDictionary *dict, id object, NSStrin
...
+	for (NSString *domain in [versionsByDomain allKeys])
 	{
 		ETPackageDescription *package = [_modelDescriptionRepository descriptionForName: domain];
 		BOOL isDeletedPackage = (package == nil);
+		int64_t version = versionsByDomain[domain] != nil
+			? (int64_t)[versionsByDomain[domain] longLongValue]
+			: (int64_t)-1;
 
-		if (isDeletedPackage || [item versionForDomain: domain] != (int64_t)package.version)
+		if (isDeletedPackage || version != (int64_t)package.version)
```
qmathe: Related to my previous comment, this isDeletedPackage case would have to be turned into an error.

qmathe:
Looks great! Thanks for reworking all this :)

Both -domainsToMigrateForItem: and -checkObject:versionForDomain: looks ok.

For the new API  +[COSSchemaMigration recordVersionsByDomain:forDomain:version:entityName:], I'm not convinced by it, I would rather pass these informations to +registerMigration:. From what I understand, it gives the initial package versions (e.g. CoreObject and Test versions) to initiate a migration, since you cannot get these initial versions from COItem anymore.
If you have version 0, 1, 2 for Test, a migration can be initiated with items whose version are set to 0, 1 or 2. For each potential initial Test version, the corresponding CoreObject version is needed, to look up the first COSchemaMigration for the CoreObject migration phase. This means +recordVersionsByDomain:forDomain:version:entityName: looks fragile. For every entity and  registered migration combination, you must call it (if we forget a single entity/call, it can break badly). It's  an extra operation you must never forget when you call +registerMigration:, so it would better to pass this information with the latter method. 

I cannot figure out why you pass the entity name to +recordVersionsByDomain:forDomain:version:entityName:. It seems this information shouldn't be needed. For starting the migration, I would say we need the the package dependency infos at every possible package version, in a data structure similar to [COSchemaMigration dependencies] which provides these same infos but only when there is a move involved.

Here is what I suggest, I would change versionsByDomainByEntityTuple to versionsByDomainByDomainVersionPair (or some other name), add a property COSchemaMigration.sourceVersionsByDomain, and in +registerMigration: populate +versionsByDomainByDomainVersionPair by extracting the same infos from the COSchemaMigration objects. This would much cleaner I think.
In fact, since COSchemaMigration.sourceVersion is read-only, COSchemaMigration.destinationVersionsByDomain would be a better choice than COSchemaMigration.sourceVersionsByDomain. This would look like:

@class COSchemaMigration
{
    NSDictionary *_destinationVersionsByDomain;
}

-setDestinationVersion: 
{
    if (self.domain == nil)
    {
        // NOTE: To eliminate this exception, we could have some null domain marker in _destinationVersions set in the initializer
        [NSException raise: NSInternalInconsistencyException format: @"Domain must not be nil to set a destination version"];
    }
    _destinationVersionsByDomain[self.domain] = @(version)
}

-setDomain:
{
    _domain = newDomain
    [_destinationVersionsByDomain replaceKey: _domain byKey: newDomain]
}

@class COSchemaMigrationDriver

-versionsByDomainForItem:
{
    versionsByDomainByEntityTuple = [COSchemaMigration versionsByDomainsByDomainVersionPair]
    key = @[item.packageName, @(item.entityVersion)]
    return versionsByDomainByDomainVersionPair[key]
}

ericwa:
I just pushed a change to update the COItem attribute names to kCOObjectEntityNameProperty, kCOObjectEntityVersionProperty, kCOObjectPackageNameProperty.
 
>I cannot figure out why you pass the entity name to +recordVersionsByDomain:forDomain:version:entityName:. It seems this information shouldn't be needed. 

I did this because, when you pass an item to -domainsToMigrateForItem:, that method needs to walk up the entity's inheritance chain and get the (package, version) pair for each superentity, to check if that (package, version) need to be migrated.

> For starting the migration, I would say we need the the package dependency infos at every possible package version

Maybe I'm missing something, but I don't think we have package dependencies, because each entity in a package can have a superentity in a different package. So there's no concept of "the parent of (pkgA, v1) is (pkgB, v0)".


Btw, I just realized a limitation of eliminating the COItem version/domain arrays. suppose there is a TypewriterDocument, v0 with superclass EtoileDocument, v0. Suppose both packages have migrations to v1.

If you open the TypewriterDocument in the generic object manager, which has the EtoileDocument metamodel and migrations (but lacks the TypewriterDocument metamodel and migrations), with the COItem version/domain arrays gone we can't migrate just the EtoileDocument part and save it. In other words because the COItem only stores the version of the subclass, any migration you do has to migrate the subclass.

If you want to support editing something with only a superclass's metamodel (EtoileDocument), and allow migrations, I think your original code with the array of versions/domains in COItem is the only way to do it. Does that sound accurate, and what do you think we should do?

SchemaMigration/COSchemaMigrationDriver.m
ericwa: Ok, I was confused because the variable is called `isDeletedEntity`.

ericwa:
I implemented something close to your suggestion of eliminating recordVersionsByDomain, and adding the property `COSchemaMigration.dependentSourceVersionsByDomain`.

I think it has to be "source versions" rather than "destination", because the code that needs to call that has an item that is from the source side of the migration.

Also didn't bother to merge this data into a special dictionary, so `versionsByDomainForItem` currently iterates through the `COSchemaMigration` objects, but that's easy to change.

PR #15, opened 2015-03-13
COSerialization: ignore unknown attributes in COItem

ericwa:
-[COObject(COSerialization) setStoreItem:]: Change from exception to NSLog warning when encountering a key in the COItem that's not in the metamodel. Test included.

An example of why this is helpful: if we had this from the start, the recent JSON change (adding a format tag) would be readable by old CoreObject versions. Also, the current schema migration design for what constitutes a breaking change depends on this.


qmathe:
Looks good. Thanks for the explanation.

PR #16, opened 2015-03-30
Schema migration

ericwa:


Core/COObjectGraphContext.m
```diff
@@ -468,8 +486,11 @@ - (void)addItemsFromItemGraph: (id <COItemGraph>)itemGraph
 	// We could also change -[COObjectGraphContext itemForUUID:] to search
 	// itemGraph during the loading rather than the loaded objects (but that's
 	// roughly the same than we do currently).
-	_loadingItemGraph = itemGraph;
-	
+	// FIXME: Update loadableUUIDs when items are deleted in the migration.
```
qmathe: This comment can be removed imo. For itemUUIDs, we don't need to substract the UUIDs of the items deleted in the migration, since _loadingItemGraph is based on the migrated items and -mainItemsFromItemGraph:loadableUUIDs: will skip UUIDs not corresponding to any item in this graph.

SchemaMigration/COSchemaMigration.h
```diff
@@ -0,0 +1,417 @@
...
+ * migration.destinationVersion = 3
+ * migration.migrationBlock = ^(COSchemaMigration *migration, NSArray *storeItems)
+ * {
+ *     NSMutableArray *newItems = [NSMutableArray new];
+ *
+ *     for (COItem *item in storeItems)
+ *     {
+ *         newItem = [item mutableCopy];
+ *         [newItem removeValueForAttribute: @"whatever"];
+ *         [newItem setVersion: migration.destinationVersion
```
qmathe: This example needs to be updated to use the latest COItem API.

One of the thing we could consider too, is changing the migration driver to update the version automatically in an additional loop just before calling -[COSchemaMigration migrateItems:] (which in turn calls the migration block here).
Do you think this would add some substantial overhead? Could it be better to let this responsability to the framework user?

SchemaMigration/COSchemaMigration.h
```diff
@@ -0,0 +1,417 @@
...
+ * migration.packageName = @"com.company.Contact"
+ * migration.destinationVersion = 3
+ * migration.migrationBlock = ^(COSchemaMigration *migration, NSArray *storeItems)
+ * {
+ *     NSMutableArray *newItems = [NSMutableArray new];
+ *
+ *     for (COItem *item in storeItems)
+ *     {
+ *         COItem *newItem = [item mutableCopy];
+ *         [newItem setVersion: migration.destinationVersion
```
qmathe: This line needs to be updated too, in order to use the latest COItem API.

SchemaMigration/COSchemaMigration.h
```diff
@@ -0,0 +1,417 @@
...
+ */
+@property (nonatomic, copy) COMigrationBlock migrationBlock;
+/**
+ * Domains depended on by the -sourceVersion of -packageName, along with
+ * their versions.
+ *
+ * This acts as a snapshot of the necessary parts of the metamodel at 
+ * -sourceVersion.
+ */
+@property (nonatomic, copy) NSDictionary *dependentSourceVersionsByDomain;
```
qmathe: This property should probably be renamed to 'dependentSourceVersionsByPackageName', and the doc updated to 'Package name depended on …'

SchemaMigration/COSchemaMigration.m
```diff
@@ -0,0 +1,139 @@
...
+}
+
+#pragma mark Migration Process -
+
+- (void)validateItems: (NSArray *)storeItems
+{
+	for (COItem *item in storeItems)
+	{
+		ETAssert(self.migrationDriver != nil);
+		NSDictionary *versionsByDomain = [self.migrationDriver versionsByDomainForItem: item];
```
qmathe: versionsByDomain should probably be renamed to versionsByPackageName

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+
+- (NSSet *)packagesToMigrateForItem: (COItem *)item
+{
+	ETEntityDescription *entity = [_modelDescriptionRepository descriptionForName: item.entityName];
+	BOOL isDeletedEntity = (entity == nil);
+	
+	/* For a deleted entity, the package versions match between item and packages */
+	if (isDeletedEntity)
+	{
+		// FIXME: We hit this line when loading a CODictionary during the test suite,
```
qmathe: This FIXME can be removed, or turned into a NOTE explaining there is no CODictionary registered as an entity description, so it's expected we hit this line on this entity name.

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+	}
+	[value addObject: object];
+}
+
+/**
+ * Returns some info on the metamodel state that the item package/version
+ * comes from, namely the package/version pairs for the entity and all of its
+ * superclasses.
+ */
+- (NSDictionary *) versionsByDomainForItem: (COItem *)item
```
qmathe: Should be renamed -versionsByPackageNameForItem:

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+{
+	for (COSchemaMigration *migration in [COSchemaMigration migrations])
+	{
+		if ([item.packageName isEqual: migration.packageName]
+			&& item.packageVersion == migration.sourceVersion)
+		{
+			NSDictionary *versionsByDomain = [migration.dependentSourceVersionsByDomain
```
qmathe: Same for this variable, which could be renamed versionsByPackageName

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+											  dictionaryByAddingEntriesFromDictionary:
```
qmathe: The indentation of these two lines look weird. May be it's just Github displaying it in a weird way though.

Core/COObjectGraphContext.m
ericwa: OK, deleted that

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+	   in the model description repository. In that case we have no migration to do */
+	if (entity.owner != nil
+		&& entity.owner.version == item.packageVersion)
+	{
+		return [NSSet set];
+	}
+	
+	/* At this point we are doing a migration for some packages for sure. */
+
+	NSDictionary *versionsByDomain = [self versionsByDomainForItem: item];
```
qmathe: The variable should probably be versionsByPackageName

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+	
+	if (versionsByDomain == nil
+		|| versionsByDomain[@"org.etoile-project.CoreObject"] == nil
```
qmathe: May be we should have a private or public constant for this CoreObject package name.

SchemaMigration/COSchemaMigration.h
ericwa: Thanks, missed that (and some other occurrences of "Domain")

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+		|| ![versionsByDomain[item.packageName] isEqual: @(item.packageVersion)])
+	{
+		// TODO: Test that we get this exception when needed
+		[NSException raise: NSInternalInconsistencyException
```
qmathe: Perhaps we should put this check in a dedicated since it's pretty big. 
Something like -validateVersionsByPackageName:compatibleWithItem:

SchemaMigration/COSchemaMigrationDriver.m
ericwa: Ok, done

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+	itemsToMigrate = [NSMutableDictionary new];
+	NSArray *upToDateItems = [self prepareMigrationForItems: storeItems];
+
+	for (NSString *packageName in [itemsToMigrate allKeys])
+	{
+		ETPackageDescription *package = [_modelDescriptionRepository descriptionForName: packageName];
+		ETAssert(package != nil);
+
+		// NOTE: Or -migrateItems:boundToPackageNamed:inRepository:
+		[self migrateItemsInDomain: packageName
```
qmathe: The NOTE should be removed, and the method -migrateItemsInDomain:toVersion: probably renamed to something like -migrateItemsBoundToPackageNamed:toVersion:

I was thinking of using BoundToPackageNamed: rather than InPackageNamed:, to make clearer the item can be processed as part of multiple packages in the migration process: the package owning it, plus packages on which this package depends on.

Tests/SchemaMigration/TestSchemaMigration.m
```diff
@@ -0,0 +1,1142 @@
...
+}
+
+- (void)testInsertOrUpdateItemsWithoutMigration
+{
+	COObjectGraphContext *context = parent.objectGraphContext;
+	
+	UKDoesNotRaiseException([context insertOrUpdateItems: [context items]]);
+}
+
+/*	- (void) testInsertOrUpdateItemsWithMigration
```
qmathe: Not sure, we should write a test like that. The scope seems to wide. What do you think?

StorageDataModel/COItem.m
```diff
@@ -447,6 +469,27 @@ - (void)removeValueForAttribute: (NSString*)anAttribute
...
+- (void)setPackageVersion:(int64_t)entityVersion
+{
+	[self setValue: @(entityVersion)
+	  forAttribute: kCOObjectPackageVersionProperty
+			  type: kCOTypeInt64];
+}
+
+- (void)setPackageName:(NSString *)packageName
+{
+	[self setValue: [packageName copy]
```
qmathe: If we copy the package name, we should probably do the same for the entity name. It doesn't really matter though.

SchemaMigration/COSchemaMigrationDriver.m
```diff
@@ -0,0 +1,428 @@
...
+			COMutableItem *combinedItem = combinedItems[item.UUID];
+			
+			if (combinedItem == nil)
+			{
+				combinedItem = pristineMutableItemFrom(item);
+				combinedItems[item.UUID] = combinedItem;
+			}
+			
+			copyAttributesFromItemTo(attributes, item, combinedItem);
+			//[combinedItem setVersion: package.version
```
qmathe: Here I think we miss a bit of code that sets the version correctly based on the final destination version for the package owning the item.
I guess this is not caught by the test suite currently. We probably a multi-steps scenario where one package containing super-entities is incremented lower than the owning one, or some other one akin to this.

I'm a bit sleepy but I think something like below should do the trick:

BOOL isOwningPackage = [packageName isEqualToString: package.name];

if (isOwningPackage)
{
   [combinedItem setVersion: package.version forPackageName: packageName];
}

SchemaMigration/COSchemaMigration.h
ericwa: Maybe we should leave it to the migration block for now, otherwise we'd have to make another copy of each item to set the version.

Tests/SchemaMigration/TestSchemaMigration.m
ericwa: Probably not.. I'm not totally sure what that test was going to be

StorageDataModel/COItem.m
ericwa: I'll add the copy, it's a little safer

SchemaMigration/COSchemaMigrationDriver.m
ericwa: Do you mind if I leave that for you? I don't totally understand this part.

qmathe: ok, I'll do it.

SchemaMigration/COSchemaMigration.h
qmathe: Yes, it's probably a bit expansive to have to make an extra copy. We can leave it like that for now.

Tests/SchemaMigration/TestSchemaMigration.m
qmathe: Basically just testing we get no exception when we are inserting or updating items that require a migration, and the resulting inner objects are identical to the one we would get by loading the items from the store, as we do in the other test methods.

qmathe:
Hey Eric,

I'd like to merge this branch into master. I fixed all the remaining issues with the latest commit a month ago or so.

For backward compatibility, we might want to relax the new exceptions I added in -[COSerialization setStoreItem:]. For example, if the item has no package name and version, don't check anything. If it has,  check both package name and version match the metamodel currently in use. If this sounds fine, I'll tweak it in this way.

Any other comments?

ericwa:
Sounds good to me. Do you mind doing a quick test that Typewriter isn't broken before merging? i.e. just that you can create a note, make an edit, close and restart with no exceptions.

ericwa:
Or I can test it once you make the change you mentioned.

qmathe:
I committed the changes I was discussing and did some testing for Typewriter, everything seems to run fine.

With Typewriter from the schema migration branch, I tested creating tags and notes, editing and navigating the history. I tested these actions with two different stores:

- one created and edited with the master, then opened and edited with the schema migration branch
- one created and edited with the schema migration branch only

If everything looks ok to you, I'll merge the changes tonight.
I tried quitting and relaunching the application few times in both cases without any issue.

ericwa:
Great, thanks for testing typewriter! Feel free to merge it.

PR #18, opened 2015-06-19
Hidden dead cross persistent root refs

qmathe:
Hey Eric,

Here are my latest changes to support both persistent root and branch deletion/undeletion with cross persistent root references.

--

The main changes are 

- updated COPrimitiveCollection concrete classes to track live vs dead references

- COCrossPersistentRootDeadRelationshipCache to track the relationships to be fixed when inner objects are undeleted (on the other side of the relationship)

- a new method -[COEditingContext updateCrossPersistentRootReferencesToPersistentRoot:Branch:isDeleted:] to call -[COObjectGraph replaceObject:withObject:] on explicit deletion/undeletion (e.g. -deletePersistentRoot:) and implicit deletion/undeletion coming through store change notifications (either for an undo track manipulation written directly to the store, or other external changes from another editing context)

- a rewritten -replaceReferencesToObjectIdenticalTo:withObject:

- COSerialization adjustments to return COPath markers for dead references, and serialize both live and dead references

- tests to cover the previous changes and few other issues

--

The current limitations are 

- -updateCrossPersistentRootReferencesToPersistentRoot:Branch:isDeleted: is currently unfaulting object graphs a bit aggressively, this can be improved later I think

- -[COMutableArray setArray:] should try to keep the dead references close to their initial position, based on some merge strategy, rather than putting them all at the end

- -[COObject replaceContentOfCollection: withCollection:propertyDescription:] is probably a useless change, I'd probably change it back to -mutableCollectionWithCollection:propertyDescription: and move the logic in -[COMutableArray setArray:] to a custom initializer (to prevent the user to accidentally inject dead references from one array into another one with the public API)

- cross persistent root keyed relationships are unfinished and untested (COMutableDictionary is ready but the rest isn't), it's not a high priority for me currently

- ivar-backed univalued relationships are unsupported currently, and more generally ivar-backed relationships are not covered by the tests ((I'll fix all that at some point) 

That's it :) Let me know if you have any questions.

Quentin.

### DIFF
--- a/Core/COEditingContext.m
+++ b/Core/COEditingContext.m
@@ -303,15 +303,8 @@
         [_persistentRootsPendingDeletion addObject: aPersistentRoot];
     }
 
-	for (COPersistentRoot *persistentRoot in [_loadedPersistentRoots objectEnumerator])
-	{
-		if (persistentRoot == aPersistentRoot)
-			continue;
-
-		// TODO: Fix references in other branches too
-		[persistentRoot.objectGraphContext replaceObject: aPersistentRoot.rootObject
-		                                      withObject: nil];
-	}
+	[self updateCrossPersistentRootReferencesToPersistentRoot: aPersistentRoot
+	                                                isDeleted: YES];
 }
 
 - (void)undeletePersistentRoot: (COPersistentRoot *)aPersistentRoot
@@ -324,18 +317,9 @@
     {
         [_persistentRootsPendingUndeletion addObject: aPersistentRoot];
     }
-	
-	/* Fix dead references in referring persistent roots */
-	
-	for (COPersistentRoot *persistentRoot in [_loadedPersistentRoots objectEnumerator])
-	{
-		if (persistentRoot == aPersistentRoot)
-			continue;
-				
-		// TODO: Fix references in other branches too
-		[persistentRoot.objectGraphContext replaceObject: nil
-											  withObject: aPersistentRoot.rootObject];
-	}
+
+	[self updateCrossPersistentRootReferencesToPersistentRoot: aPersistentRoot
+	                                                isDeleted: NO];
 	
 	/* Once all the dead references are fixed, remove them from the cache */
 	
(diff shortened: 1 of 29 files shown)


Core/COEditingContext.m
```diff
@@ -310,6 +325,105 @@ - (void)undeletePersistentRoot: (COPersistentRoot *)aPersistentRoot
...
+	NSParameterAssert(aBranch == nil
+		|| (!aBranch.isCurrentBranch && aBranch.persistentRoot == aPersistentRoot));
+
+	for (COPersistentRoot *persistentRoot in [_loadedPersistentRoots objectEnumerator])
+	{
+		if (persistentRoot == aPersistentRoot)
+			continue;
+		
+		// TODO: Use -objectGraphWithoutUnfaulting or -allObjectGraphContexts to
+		// prevent loading every object graph contexts we check
```
ericwa: Yeah, this TODO would be good to have.

From what I can see, this method is a "traverse loaded object graphs and fix up pointers" kind, it really shouldn't do any unfaulting, for both performance and correctness.

Also, it seems to be fixing all loaded persistent roots, couldn't it use the dead reference cache/relationship cache to only fix up ones that need it?

ericwa:
Thanks, this looks nice. I enabled the tests for dead cross refs I wrote last year in TestCrossPersistentRootReferences.m, they all work now!

I haven't read the diff in detail but the concept sounds fine and the tests look good.
Just had the one comment above.

I think this should be fine to merge in though. Any performance concerns? any potential compatibility problems with existing stores?

Core/COEditingContext.m
qmathe: Initially I wrote it to unfault all object graphs to make tests easier to write and truly exhaustive. If we don't unfault all object graphs requiring reference fixes in -updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted:, we must do it in the tests themselves to ensure they remain exhaustive. That's what we should truly do.

To reply to your last question, the code checks every branch (in every persistent root) to fix references. In -[COObjectGraphContext replaceObject:withObject:], I check dead reference cache/relationship cache and return immediately when the object graph is not concerned, otherwise I call -[COObject replaceReferencesToObjectIdenticalTo:withObject:]. 
As you suggest, it is surely possible to move this check to -updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted:. I haven't really thought about it, since the most expansive part is to iterate over all the inner objects and we skip this one already. The other thing is that I would wait until lazy loading of persistent roots is implemented, since this could require other changes to the current implementation. Once we support lazy loading, iterating over all branches should be less costly too.
This puts aside, moving the check to -updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted: would make the code simpler and easier to debug too.

qmathe:
Another things I didn't do is to refactor TestCrossPersistentRootReferences.m. I thought we should discuss together how we want to reorganize it. We can probably remove some tests in it, that are now covered in more focused and exhaustive way by the new tests I wrote. Some other tests should probably be broken into smaller ones and move to the new test classes I created too. 
Few more complex test scenarios (e.g. the one that involve person names and based on a Keynote document) could remain in TestCrossPersistentRootReferences.

The changes should be fully compatible with existing stores. 

Concerning performance issues, I don't feel there is a difference when running the test suite or playing with my iOS app. I'll check the Benchmark results before merging though. 

To be more precise, we should have no additional costs when no persistent root / branch deletion/undeletion is triggered explicitly or implicitly (e.g. by going back in the past). The dead relationship cache should add zero overhead.  There are two issues to be corrected in the long run though:

- COMutableArray is much slower for big arrays on some operations. Every time there is a call to -backingIndex: (e.g. with -objectAtIndex:) it enumerates the indexes to find the matching one since NSIndexSet is not ordered. To fix this issue, we would need a dedicated COIntArray or COOrderedIndexSet. It's worth to mention that the issue doesn't exist during serialization/deserialization since the backing is accessed directly without having to check dead vs live indexes.
- -updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted: is iterating over every persistent roots and branches on each call as you pointed it out. It could be pretty costly when 10000 persistent roots or more are involved… For example, you go back in the past and 10000 persistent roots have changed or you delete 10000 persistent roots with -[COPersistentRoot setDeleted:]. When we receive a store change notification, may be it's possible to do everything in a single loop, but if optimizing it as you suggested is enough, I wouldn't risk this extra complexity.

Any special changes or checks I should do before merging it? 
I'll test Typewriter run correctly with it.

Core/COEditingContext.m
ericwa: OK thanks for the explanation. Let's leave it as a to-do item then

ericwa:
I think it's fine to merge, if any issues come up with typewriter we can fix them

qmathe:
ok, I just merged it. Thanks for the review btw.

Before merging it, I did some testing with Typewriter and found some bugs I fixed. Can you try Typewriter and let me know if it runs fine for you?

For the Benchmark target, the results are unchanged.

ericwa:
Thanks. Only tested it briefly, but Typewriter seems fine, will watch out for any issues with it.

Issue #21, opened 2015-06-20
No way to reload COUndoTrackStore


qmathe:
I'm looking for a way to reload COUndoTrackStore at run-time. The goal is to import a backup that consists of a CoreObject store plus the undo track store that goes along with it, and doing so while my application is running.

Something like -[COUndoTrackStore reload] (similar to COUndoTrackStore initializer) could be good enough. -[COUndoTrackStore reload] could post a change notification telling the undo tracks to reload their content for the new undo track store.

This would be a private API though. Not sure if we should add public API like +[COUndoTrack reloadUndoTrackStore] or similar.

ericwa:
Iirc, should be straightforward. 

It would be dangerous to replace the sqlite db files with other ones while CoreObject is running. But if you can be sure all sqlite db connections are closed, should be safe. (But then you wouldn't need a reload method?)

Might be better to have a -export and -import: methods that write/read json.

I'm just on my phone at the moment so will take a look at the code later.

PR #22, opened 2015-07-08
Undo track store optional singleton

qmathe:
For the issue #21, I gave some more thoughts to the problem. I also hit some other related issues while working on the history compaction, which motivated me to work out a cleaner approach.

The core problem is that the undo track store is a singleton. This works nicely when applications (using the same CoreObject version) want to use a common CoreObject store and share their histories, but in other more sandboxed situations, it can be troublesome. 

So I made some changes to make the singleton optional, the framework user can now provide a custom undo track store to the editing context. The undo tracks now request this store to their editing context. When using a custom undo track store, releasing editing contexts and their undo tracks will now release all CoreObject state in the current process. 

The main use cases are:

- applications that don't want to mix their history with other applications
- swapping stores at run-time (for backup support)
- using incompatible CoreObject versions on the same system
- testing undo tracks with a pristine state
- avoiding mixing test histories with real application histories

The last two points are problems I encoundered while working on history compaction.
The second point is related to issue #21.

The conceptual model and current limitations are documented in COUndoTrackStore class description.

btw, I think I figured out the issues about opened vs closed connections we had. Basically when an exception is raised, between the frame where where an exception is thrown and the one where it's caught, all arguments that got retained by ARC won't get released.

### DIFF
--- a/Core/COEditingContext.h
+++ b/Core/COEditingContext.h
@@ -178,8 +178,10 @@
  * To register types bound to classes manually, see
  * -[ETModelDescriptionRepository registerEntityDescriptionsForClasses:resolveNow:].
  *
- * For a nil model repository, or a repository that doesn't a COObject entity 
+ * For a nil model repository, or a repository that doesn't a COObject entity
  * description, raises a NSInvalidArgumentException.
+ *
+ * For a nil undo track store, raises a NSInvalidArgumentException.
  */
 - (instancetype)initWithStore: (COSQLiteStore *)store
    modelDescriptionRepository: (ETModelDescriptionRepository *)aRepo
--- a/Core/COEditingContext.h
+++ b/Core/COEditingContext.h
@@ -10,7 +10,8 @@
 #import <CoreObject/COPersistentObjectContext.h>
 
 @class COSQLiteStore, COEditingContext, COPersistentRoot, COBranch, COObjectGraphContext, COObject;
-@class COUndoTrack, COCommandGroup, COCrossPersistentRootDeadRelationshipCache, CORevisionCache;
+@class COUndoTrackStore, COUndoTrack, COCommandGroup;
+@class COCrossPersistentRootDeadRelationshipCache, CORevisionCache;
 @class COError;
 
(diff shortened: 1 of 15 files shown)


ericwa:
This looks good! feel free to merge

btw, were we really forgetting to delete the undo track store between test cases? thanks for noticing that.

Also

> basically when an exception is raised, between the frame where where an exception is thrown and the one where it's caught, all arguments that got retained by ARC won't get released.

isn't that a bug in ARC?

qmathe:
Thanks for taking a look. I just merged the changes into master. 

We were not deleting the undo track store between test cases, since the tests were using +[COUndoTrackStore defaultStore] which returns the one located in Library/CoreObject/Undo/undo.sqlite (the same one Typewriter used). If we were deleting it between tests, Typewriter undo history would have disappeared every time we ran the test suite.

For the exception leak issue, looks like it's a documented ARC limitation: http://clang.llvm.org/docs/AutomaticReferenceCounting.html#exceptions

It seems this flag -fobjc-arc-exceptions can eliminate these leaks, however it isn't passed by default probably because manual reference counting suffers from similar limitations (it involves potentially never executing release calls following the line were the exception is raised).

In our case, do you know if all the code is compiled as ObjC++ or only .mm files? For ObjC++, -fobjc-arc-exceptions is implicitly defined.

I tried passing -fobjc-arc-exceptions as a C flag with Xcode, but the 3 opened databases remain the same. So I'm a bit puzzled.

If we can make it work properly, may be we should pass this flag just for the test suite? … or keep it as is to maintain the same exact memory management everywhere (in the test suite and CoreObject applications).

For some more background about this exception leak, you can check this link too: http://stackoverflow.com/questions/4648952/objective-c-exceptions/4649234#4649234

ericwa:
Thanks for the info. I expect only the .mm files are compiled as ObjC++.

IIRC, the exceptions only happen in corner cases when the store is deleted, etc. I don't remember putting too much care in to those cases, so an exception + leak is probably ok for now.

PR #23, opened 2015-08-27
History compaction

ericwa:


Core/COBranch.m
```diff
@@ -1045,7 +1049,6 @@ - (BOOL)setCurrentNode: (id <COTrackNode>)node
 {
 	INVALIDARG_EXCEPTION_TEST(node, [node isKindOfClass: [CORevision class]]);
 	[self setCurrentRevision: (CORevision *)node];
-	[self didUpdate];
```
ericwa: why remove this -didUpdate call?

Core/COEditingContext.h
```diff
@@ -15,6 +15,27 @@
...
+	 * on deletion.
+	 */
+	COEditingContextUnloadingBehaviorNever,
+	/**
+	 * Persistent roots can be unloaded on deletion.
+	 *
+     * For external deletions committed in other editing contexts, persistent
+	 * roots will unloaded in the current one.
+	 */
+	COEditingContextUnloadingBehaviorOnDeletion
```
ericwa: IMO, COEditingContextUnloadingBehaviorOnDeletion could be the only supported option. 

The design we have when a COEditingContext receives a notification that some changes were made in the store is, the changes take effect immediately (branch switch, current revision of branch switch, new persistent root, etc.), and there's no way to opt out of this. The rationale is, everything is undoable anyway, so it would just add complexity to have fine-grained control of propagating changes from store->editing context.

Core/COBranch.m
qmathe: It hasn't been truly removed but moved to -setCurrentRevisionSkipSupportsRevertCheck:, so the notifications are properly posted even when -setCurrentRevision: is called. For example, -undo calls -setCurrentRevision:.

Core/COEditingContext.h
qmathe: I agree.

COEditingContextUnloadingBehaviorNever matters for testing. We can make it private for now.

In future, we can make it public again if we want to expose new options such as COEditingContextUnloadingBehaviorOnMemoryPressure/OnMaxCount. Or if it is helpful for debugging client code.

Undo/COUndoTrackStore.m
```diff
@@ -190,81 +181,209 @@ - (instancetype)init
...
+		dispatch_semaphore_signal(_transactionLock);
+	}
+	return ok;
 }
 
 - (BOOL) commitTransaction
 {
+	ETAssert([NSThread isMainThread]);
+
     BOOL ok = [_db commit];
```
ericwa: -commit sends a regular SQL statement, it should be done in the `_queue`, I think.
While SQLIte is supposed to be threadsafe by default (see https://www.sqlite.org/threadsafe.html ) I don't think fmdb is.

Undo/COUndoTrackStore.m
```diff
@@ -190,81 +181,209 @@ - (instancetype)init
...
 - (BOOL) beginTransaction
 {
-    return [_db executeUpdate: @"BEGIN EXCLUSIVE TRANSACTION"];
+	ETAssert([NSThread isMainThread]);
+
+	// If there is a background operation (e.g. mark as deleted, vacuum) underway,
+	// wait until it is finished
+	dispatch_semaphore_wait(_transactionLock, DISPATCH_TIME_FOREVER);
+
+    BOOL ok = [_db beginTransaction];
```
ericwa: Should be done in _queue

Undo/COUndoTrackStore.m
```diff
@@ -283,55 +402,180 @@ - (id) deserialize: (NSData *)data
...
+ * We run vacuum with dispatch_sync() in a queue, so we can sure no other
+ * threads access the database when we invoke -vacuum in a background thread.
+ */
+- (BOOL)vacuum
+{
+    assert(dispatch_get_current_queue() != _queue);
+	__block BOOL success = NO;
+
+	// If there is a transaction underway, wait until it is finished
+	dispatch_semaphore_wait(_transactionLock, DISPATCH_TIME_FOREVER);
```
ericwa: I don't understand why _transactionLock is needed (once the begin/commit transaction calls are moved to _queue as I suggested above).

Just to confirm, the addition of `_queue` is just for future use and doesn't actually provide any concurrency right now, correct?

Once all DB access is moved to `_queue`, it should be safe to make -vacuum use `dispatch_async` instead of `dispatch_sync`. This may not actually help because the next thing to call `dispatch_sync` will block until the vacuum finishes.

Undo/COUndoTrackStore.m
```diff
@@ -190,81 +181,209 @@ - (instancetype)init
 - (void) dealloc
 {
     [_db close];
```
ericwa: do in `dispatch_sync(_queue,`

Undo/COUndoTrackStore.m
```diff
@@ -190,81 +181,209 @@ - (instancetype)init
...
+	// currently (we compile CoreObject with -DOS_OBJECT_USE_OBJC=0).
+	dispatch_release(_queue);
+	dispatch_release(_transactionLock);
+#endif
+}
+
+- (BOOL) setupSchema
+{
+    /* SQLite compatibility */
+
```
ericwa: do in `dispatch_sync(_queue,`

Undo/COUndoTrackHistoryCompaction.h
```diff
@@ -0,0 +1,89 @@
...
+ * @section Common Use Cases
+ *
+ * The most common use case is when you want to free space in a CoreObject store.
+ *
+ * Before passing the compaction strategy to -[COSQLiteStore compactHistory:], 
+ * you must always call -compute.
+ *
+ * For now, COUndoTrackHistoryCompaction is not thread-safe.
+ */
+@interface COUndoTrackHistoryCompaction : NSObject <COHistoryCompaction>
```
ericwa: So I understand this correctly, is the idea here that this object lets a COUndoTrack control the compaction of a COSQLiteStore, such that.. in GC terms, the undo track keeps alive everything that it references in the store?

I was expecting that the user would control compaction by looking at the revision graph of persistent roots, but if you aren't exposing those in a UI , and only exposing undo tracks, this makes sense I guess.

Undo/COUndoTrackStore.m
qmathe: ok, it's corrected now.

Undo/COUndoTrackStore.m
qmathe: Fixed

Undo/COUndoTrackStore.m
qmathe: -setupSchema is called in the database queue already (if you look at the initializer), so I just added an assertion that prevents the method to be called from another queue.

Undo/COUndoTrackStore.m
qmathe: Fixed

Undo/COUndoTrackStore.m
qmathe: _transactionLock prevents the following sequence to be created in the database queue:

- begin transaction
- vacuum (or -markCommandsAsDeleted)
- end transaction (aka commit)

Since a COUndoTrackStore transaction is not submitted as a single block to the database queue, it seems to me we must prevent incompatible operations (e.g. vacuum or mark as deleted) to be submitted before this transaction ends.

_queue is used to run -vacuum in background (and blocks any other COUndoTrackStore operations that get submitted to the queue) while keeping the main thread available present a spinning indicator.
Beside this, there is no real concurrency use with _queue currently.

dispatch_async() doesn't seem compatible with -vacuum, since no other SQLite operations must be run during a vacuum, so I think we want to block the database queue during a vacuum operation. Am I missing something?

ericwa: Ah, I missed the fact that you want to prevent calling -vacuum during a transaction.
However, I still think a semaphore is unnecessary because we don't need to check the lock from multiple threads.

What about adding a BOOL _inTransaction ivar?
```
-beingTransaction
{
    ETAssert(is main thread);
    _inTransaction = YES;
    dispatch_async(...); // enqueues SQL commands
}
-endTransaction
{
    ETAssert(is main thread);
    ETAssert(_inTransaction);
    _inTransaction = NO;
    dispatch_async(...); // enqueues SQL commands
}
-vacuum
{
    ETAssert(is main thread);
    ETAssert(!_inTransaction);
    dispatch_async(...); // enqueues SQL commands
}
````

Re:

    dispatch_async() doesn't seem compatible with -vacuum, since no other SQLite operations must be run during a vacuum, so I think we want to block the database queue during a vacuum operation. Am I missing something?

AFAIK, dispatch_async() just means the calling thread doesn't block. The block is added to the queue and runs in a background thread, just like with dispatch_sync. So we never have multiple SQLite operations running at once; the blocks in the dispatch queue always execute serially (whether dispatch_async or dispatch_sync is used.) So I don't see why dispatch_async would be incompatible with -vacuum.

Undo/COUndoTrackHistoryCompaction.h
qmathe: Yes, the idea is that the undo track can keep anything it references alive:

- old revisions
- deleted persistent roots
- deleted branches

Non-deleted persistent roots and non-deleted branches won't be finalized, even when they are not referenced by an undo track.

In my latest application iteration, I only expose undo tracks. The history compaction is driven by a pattern track that match all tracks in the undo track store.

Previously I was exposing some current branches directly in the UI (CORevision presented as a COTrackNode), but I was still recording every commit on undo tracks, so the GC targeting the revisions was able to work properly without breaking the branch revision graphs at the UI level (e.g. no missing revisions in the middle of a branch history or no revisions unrelated to the undo track nodes kept alive).

Undo/COUndoTrackStore.m
qmathe: ok, I understand what you mean. 

I did use a semaphore rather than a boolean to prevent race conditions, since I was expecting both -vacuum and -begin/endTransaction to be called on different threads. AFAIK, when two threads runs on different cores, accessing the same boolean variable from both is not safe, since the value is not garanteed to be updated exactly at the same time on both cores.

If we use dispatch_async() instead of dispatch_sync() then the problem disappears since the boolean variable would always be accessed from the main thread.

I choose to use dispatch_sync() since it makes easy to do stuff after -vacuum without an extra completion hander. For example:

dispatch_async(someBackgroundQueue, {
    [store vacuum];
    // Add some code here to notify that the vacuum is done (-vacuum is done at this point when using dispatch_sync() inside its implementation)
})

Now if we change the API to something like -vacuumWithCompletionHandler: then using dispatch_async() inside it should work fine. For example:

[store vacuumWithCompletionHandler: {
    // Add some code here to notify that the vacuum is done
}]

I guess this leads to a simpler and mor explicit API and lets us eliminate the extra semaphore. What do you think?

In this case, I think all methods (e.g. -markCommandsAsDeleted) which support to be invoked in background should take a completion handler argument.

ericwa: Ah ok, I understand now :-). My preference is for the `-vacuumWithCompletionHandler:` API. AFAIK the whole public API of CoreObject is non-threadsafe, and imo `-vacuumWithCompletionHandler:` is safer/more explicit than having `vacuum` be a special method that is safe to call from a background thread.

Btw, I just tried running vacuum from the sqlite3 interactive shell on a Typewriter store that is about 150mb, it took around 5 seconds.

qmathe: ok, I'll switch the code to -vacuumWithCompletionHandler:.

Yes, vacuum tends be pretty fast in my experience, but not instantaneous unless you have a really small CoreObject store.

qmathe: I tried to implement -markCommandsAsDeletedForUUIDs:withCompletionHandler:, -finalizeDeletionsWithCompletionHandler and -vacuumWithCompletionHandler, but it doesn't work out, because this requires to change all calling APIs to use a completion handler.

So -[COHistoryCompaction beginCompaction] needs to be turned into -beingCompactionWithCompletionHandler:. In the same way, -[COSQLiteStore compactHistory:] needs to be rewritten into -compactHistoryWithCompletionHandler:, but then this means changing more COSQLiteStore methods and introducing some asynchronous API (e.g. -finalizeDeletionsForPersistentRoot:error:withCompletionHandler:). This doesn't sound like a good idea, since COSQLiteStore API is designed as an entirely synchronous API.

I now remember that I also wrapped COUndoTrackStore background operations such as vacuum with a synchronous API in order to mimick COSQLiteStore API.

So I'm not sure what to do… I'm tempted to say that we should keep both COSQLiteStore and COUndoTrackStore APIs purely synchronous (to keep the API symetric). Using completion handlers aslo translate into less readable implementation code due to the nested callbacks. 

May be we should postpone this change?

ericwa: Hmm, ok. I'm fine with merging the pull request as it is.

I agree using completion handlers everywhere leads to unreadable code (http://callbackhell.com/) ;-), and we don't want to have a cascade of API changes because the current COSQLiteStore api seems good as-is.

However, I don't think a synchronous -vacuum method designed to be called from a background thread is a good idea. It gets us into a lower level of multithreaded code, and it's hard to tell if there are subtle threading bugs with the use of a semaphore, etc.
We can leave it for now as a temporary hack, and then change it later.

Another option of -vacuum is just do dispatch_async with no completion handler.

qmathe: If we do dispatch_async() without a completion handler for -vacuum then we cannot know when the vacuum ends, right? I mean, we want to be told about it in order to remove the spinning indicator in the UI.

For -finalizeDeletions and -marksCommandsAsDeletedForUUIDs:, in a similar manner we have code that cannot be executed until these methods are done.

I'd be in favor of removing the semaphore too. A good option could be to solve the problem in the same way COSQLiteStore does it. By introducing a COUndoTrackStoreTransaction class hierarchy, we could eliminate -[COUndoTrackStore begin/endTransaction] and thereby the transaction lock too. This solution might seem a bit heavy, but it could result in a cleaner code in COUndoTrack too.

PR #24, opened 2015-09-07
COObject multivalue optimizations

ericwa:
The first 4 commits are just adding tests/benchmarks, the next 4 have the optimizations:

* rewrote handling of tombstones in `COMutableArray` to be faster. `-count` and `-objectAtIndex:` are now trivial, O(1).
* make `COUnsafeRetainedMutableArray` automatically remove duplicates, like NSOrderedSet. This lets us avoid the slow `-[COObject removeDuplicatesInValue:propertyDescription:]` every time a collection is modified.
* make a fast path version of `-commonDidChangeValueForProperty:` that just handles the updated elements of the collection. It's called: `-commonDidChangeValueForProperty:atIndexes:withObjects:mutationKind:`.

The benchmark I measured this with is TestLargeOrderedRelationsip (added in the first commit in this branch).

The benchmark incrementally adds 1000 OutlineItems to an OutlineItem with -addObject, then times various calls on `parent.contents` like `-count` and for/in enumeration.
Rewriting the tombstone code in COMutableArray made  `-count` ~30x faster when called on a collection with 1000 elements, and for/in enumeration ~5x faster.

The slowest part originally was building the collection of 1000 outline items by calling -addObject: in a loop, which was taking 10 seconds! Adding the fast path for `-commonDidChangeValueForProperty:` brought this down to 0.25 seconds (still kind of bad, but tolerable.)

(edit: all benchmarks re-run with 17aefef )
Before: 
```
2015-09-11 13:16:45.358 BenchmarkCoreObject[79618:1863789] 92.141032 us per iteration for 'Create OutlineItem and add to an NSMutableArray with -addObject:'
2015-09-11 13:16:56.758 BenchmarkCoreObject[79618:1863789] 11.399381 ms per iteration for 'Add OutlineItem to an OutlineItem with -addObject:'
2015-09-11 13:16:56.758 BenchmarkCoreObject[79618:1863789] 0.133991 us per iteration for 'Add OutlineItem to an NSMutableArray with -addObject:'
2015-09-11 13:16:56.759 BenchmarkCoreObject[79618:1863789] 0.459015 us per iteration for '-count on CoreObject array with 1000 elements'
2015-09-11 13:16:56.763 BenchmarkCoreObject[79618:1863789] 0.009000 us per iteration for '-count on NSArray with 1000 elements'
2015-09-11 13:17:00.693 BenchmarkCoreObject[79618:1863789] 3.930156 ms per iteration for 'for/in loop on CoreObject array with 1000 elements'
2015-09-11 13:17:00.698 BenchmarkCoreObject[79618:1863789] 4.337966 us per iteration for 'for/in loop on NSArray with 1000 elements'
```
After:
```
2015-09-11 13:15:34.698 BenchmarkCoreObject[79367:1862903] 155.786991 us per iteration for 'Create OutlineItem and add to an NSMutableArray with -addObject:'
2015-09-11 13:15:34.829 BenchmarkCoreObject[79367:1862903] 130.748034 us per iteration for 'Add OutlineItem to an OutlineItem with -addObject:'
2015-09-11 13:15:34.830 BenchmarkCoreObject[79367:1862903] 0.187993 us per iteration for 'Add OutlineItem to an NSMutableArray with -addObject:'
2015-09-11 13:15:34.830 BenchmarkCoreObject[79367:1862903] 0.014007 us per iteration for '-count on CoreObject array with 1000 elements'
2015-09-11 13:15:34.830 BenchmarkCoreObject[79367:1862903] 0.009000 us per iteration for '-count on NSArray with 1000 elements'
2015-09-11 13:15:34.970 BenchmarkCoreObject[79367:1862903] 139.808953 us per iteration for 'for/in loop on CoreObject array with 1000 elements'
2015-09-11 13:15:34.974 BenchmarkCoreObject[79367:1862903] 4.342973 us per iteration for 'for/in loop on NSArray with 1000 elements'
```

Edit: after removing relationship snapshotting:
```
2015-09-11 13:14:07.742 BenchmarkCoreObject[78716:1861487] 126.591027 us per iteration for 'Create OutlineItem and add to an NSMutableArray with -addObject:'
2015-09-11 13:14:07.791 BenchmarkCoreObject[78716:1861487] 48.901021 us per iteration for 'Add OutlineItem to an OutlineItem with -addObject:'
2015-09-11 13:14:07.792 BenchmarkCoreObject[78716:1861487] 0.321984 us per iteration for 'Add OutlineItem to an NSMutableArray with -addObject:'
2015-09-11 13:14:07.792 BenchmarkCoreObject[78716:1861487] 0.021994 us per iteration for '-count on CoreObject array with 1000 elements'
2015-09-11 13:14:07.793 BenchmarkCoreObject[78716:1861487] 0.009000 us per iteration for '-count on NSArray with 1000 elements'
2015-09-11 13:14:07.951 BenchmarkCoreObject[78716:1861487] 157.586992 us per iteration for 'for/in loop on CoreObject array with 1000 elements'
2015-09-11 13:14:07.955 BenchmarkCoreObject[78716:1861487] 4.539967 us per iteration for 'for/in loop on NSArray with 1000 elements'
```

### DIFF
--- a/Core/COObject.m
+++ b/Core/COObject.m
@@ -291,7 +291,7 @@ See +[NSObject typePrefix]. */
 	_isPrepared = YES;
 	_variableStorage = [self newVariableStorage];
 	_incomingRelationshipCache = [[CORelationshipCache alloc] initWithOwner: self];
-	_oldValues = [NSMutableArray new];
+	_propertyChangeStack = [NSMutableArray new];
 	_additionalStoreItemUUIDs = [self newAdditionalStoreItemUUIDs: (inserted == NO)];
 
 	[_objectGraphContext registerObject: self isNew: inserted];
@@ -367,7 +367,7 @@ See +[NSObject typePrefix]. */
 	newObject->_objectGraphContext = _objectGraphContext;
 	newObject->_variableStorage = [self newVariableStorage];
 	newObject->_incomingRelationshipCache = [[CORelationshipCache alloc] initWithOwner: self];
-	newObject->_oldValues = [NSMutableArray new];
+	newObject->_propertyChangeStack = [NSMutableArray new];
 
 	return newObject;
 }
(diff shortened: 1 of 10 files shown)


Core/COObject.m
```diff
@@ -1331,6 +1435,96 @@ - (void)commonDidChangeValueForProperty: (NSString *)key
...
+		replacedOrRemoved = @[];
+	}
+	else if (mutationKind == ETCollectionMutationKindRemoval)
+	{
+		addedObjects = @[];
+		replacedOrRemoved = objects;
+	}
+	else if (mutationKind == ETCollectionMutationKindReplacement)
+	{
+		// FIXME: How do we know which objects were replaced if it's an unordered collection?
```
ericwa: @qmathe  Wanted to point this out note I left.. it seems like a gap in the ETCollection notification API (...property:atIndexes:withObjects:mutationKind:). Or maybe ETCollectionMutationKindReplacement is only used for ordered collections?

ericwa:
Argh, my indentation looks awful in github, here is what I'm using:
screen shot 2015-09-07 at 1 51 02 am

Also, I know the last commit is ugly, I don't mind refactoring it more before merging if you want.

I think it is important to address that as soon as possible though (only processing collection changes, not re-checking all elements); I'm surprised we didn't do it before release because it only took a few hours.

Core/COObject.m
qmathe: ETCollectionMutationKindReplacement is valid for both unordered and ordered collections, but there is a gap as you pointed it out. 

KVO  notifications don't tell you directly about which objects were replaced for unordered collections, so I kept a similar gap for ETCollectionMutationKindReplacement and unordered collections. I think this gap exists in KVO notification, because you can compute it from the old vs new collection differences.

So we can probably compute the replaced objects (COObject takes a snapshot of the old collection in -willChangeValueForProperty:). The old collection is 'oldValue' and the new one 'newValue'. Something like 'replacedObjects = [oldValue minusSet: newValue]' should do the trick, unless I'm missing something.

qmathe:
Thanks for this refactoring, it looks great :)

For the indentation, I'm using the same settings than you. However I have turned off Xcode automatic indentation and I tend to indent everything by hand. For Étoilé tabs + spaces combination to indent method keywords (and keep things aligned indepedently of the tab size), this doesn't work well with Xcode which always wants to turn spaces into tabs everyhere. So I tend to do everything by hand or use some more flexible editor.

I don't mind if the indentation is somewhat broken. This can be fixed later.

Core/COObject.m
ericwa: I'd like to get rid of the snapshot of old collection values in -willChangeValueForProperty:, if possible, because it's slow. It looks like most of the time in OutlineItem -addObject: is spent creating the snapshot and deallocating it, at least when the average size of the collection is 500 items.

qmathe: I tried to avoid this extra snapshot when I wrote this code, but it's probably pretty hard to do, so it could be better to keep this change for another pull request I think.

-updateCachedOutgoingRelationshipsXXX relies on this snapshot and we have a check to delect badly paired -will/DidChangeValueXXX  in -popOldCoreObjectRelationshipValueForPropertyDescription:. This last method is pretty ugly, but debugging missing -will/DidChangeValueXXX is hard.

The really ugly thing is that Apple is already taking a snapshot of the old collection in -willChangeValueForKey: but gives no API to access it in -didChangeValueForKey: (-didChangeValueForProperty: in our case). It's only available when you receive the KVO notification in -observeValueForKeyPath:ofObject:change:context: (as NSKeyValueChangeOldKey). 
So I'm not sure what the solution would be. Writing some dedicated code to make these snapshots cheap or hacking into KVO private API, or something else?

ericwa: I was going to describe my idea for removing the snapshots but went ahead and implemented it in  e58f7b6 . What I did is just move the part of the relationship cache update that needs access to the old value. It's now moved to -commonWillChangeValueForProperty:. I also added a fast path version of -commonWillChangeValueForProperty:atIndexes:...


I kept the stack that tests for incorrect calls to -willChange../didChange..:, but just store the property name being edited there.

qmathe: This looks good and should be much more efficient than what we have.

PR #28, opened 2015-11-09
Made possible to customize presentation of implicit undo track operations

qmathe:
I made some changes to support presenting a selective undo/redo with a custom text or style synthesized from the applied (inverse) command metadata.  

We can show a button that scrolls to the applied command, or choose a better presentation than Selective Undo 'xxx' or Undo 'xxx'. 

For example, we can apply a strikethrough style to the text. If the command is undone another time, we could remove the strikethrough style and present it exactly as the original command rather than Selective Undo 'Selective Undo 'xxx''.

History recorded prior to this change won't be customizable, but will remain presented as it was previously.

For the now metadata keys are:

- kCOCommitMetadataNodeUUID
- kCOCommitMetadataNodeOperationIdentifier
- kCOCommitMetadataNodeInversed (this one is a convenience and can be computed by traversing undo operations with kCOCommitMetadataNodeUUID)

kCOCommitMetadataShortDescription and kCOCommitMetadataShortDescriptionArguments are also set to the same values than the commit metadata of kCOCommitMetadataNodeUUID. This makes it easy to interpolate them in a custom way inside -[COCommitDescription localizedShortDescription] or in client code.

For the metadata keys, may be the following would be better or more accurate:

- kCOCommitMetadataOperationCommandUUID
- kCOCommitMetadataOperationIdentifier
- kCOCommitMetadataOperationInversed

Not sure, what are the best metadata keys… what do you think?


Utilities/COCommitDescriptor.h
```diff
@@ -309,3 +309,36 @@ extern NSString *kCOCommitMetadataShortDescription;
  * See also COCommitDescription Localization documentation section.
  */
 extern NSString *kCOCommitMetadataShortDescriptionArguments;
+/**
+ * The optional key that indicates the command group UUID used to synthesize
+ * the commit.
```
ericwa: Maybe "command group UUID which this commit selectively applies or undoes" would be clearer?

ericwa: Also, what about renaming this to kCOCommitMetadataBaseNodeUUID?

Undo/COUndoTrack.m
```diff
@@ -566,13 +573,30 @@ - (void) doCommand: (COCommandGroup *)aCommand inverse: (BOOL)inverse addToStore
...
-	
+
 	NSMutableDictionary *md = [NSMutableDictionary new];
-	md[kCOCommitMetadataIdentifier] = inverse ? @"org.etoile.CoreObject.undo" : @"org.etoile.CoreObject.redo";
-	if ([aCommand localizedShortDescription] != nil)
+	NSString *identifier = aCommand.metadata[kCOCommitMetadataIdentifier];
+	NSString *shortDescription = aCommand.metadata[kCOCommitMetadataShortDescription];
+	NSArray *args = aCommand.metadata[kCOCommitMetadataShortDescriptionArguments];
+	NSNumber *inversedValue = aCommand.metadata[kCOCommitMetadataNodeInversed];
+	BOOL inversed = inverse && (inversedValue == nil || !inversedValue.boolValue);
```
ericwa: I'm not sure if checking aCommand.metadata[kCOCommitMetadataNodeInversed] is a good idea here.

e.g. if you have 3 commits: "A", "undo(A)", "undo(undo(A))", what you have implemented currently will give "undo(undo(A))" an kCOCommitMetadataNodeInversed value of NO, but the kCOCommitMetadataNodeUUID will be set to "undo(A)", which are contradictory (the last commit is not applying "undo(A)", but un-applying it). I think it would be better to just have `BOOL inversed = inverse;`. If the API user wants to, they can follow the chain of kCOCommitMetadataNodeUUID to find out that this selective undo is in fact re-applying an original commit.

Utilities/COCommitDescriptor.h
```diff
@@ -309,3 +309,36 @@ extern NSString *kCOCommitMetadataShortDescription;
...
+ * undo/redo or implicit undo due to a pattern undo track manipulation).
+ *
+ * The value is a string representing a commit descriptor identifier.
+ *
+ * When this key is set, -[COCommitDescriptor localizedShortDescription] will 
+ * synthesize a special description by interpolating both 
+ * kCOCommitMetadataShortDescription + kCOCommitMetadataShortDescriptionArguments 
+ * into the short description template for the operation identifier.
+ */
+extern NSString *kCOCommitMetadataNodeOperationIdentifier;
```
ericwa: Actually I was just reading some code above and forgot what kCOCommitMetadataNodeOperationIdentifier was.
Could we rename it to kCOCommitMetadataUndoType?

Utilities/COCommitDescriptor.h
ericwa: Edit: kCOCommitMetadataUndoBaseUUID ?

Utilities/COCommitDescriptor.h
```diff
@@ -309,3 +309,36 @@ extern NSString *kCOCommitMetadataShortDescription;
...
+/**
+ * The optional key that indicates whether the commit was created by inversing 
+ * the command referenced by kCOCommitMetadataNodeUUID.
+ *
+ * The value is a NSNumber boolean.
+ */
+extern NSString *kCOCommitMetadataNodeInversed;
```
ericwa: IMO the "Node" doesn't add much information, how about kCOCommitMetadataInversed ?

Undo/COUndoTrack.m
```diff
@@ -293,60 +293,67 @@ another command branch (the head command changes) */
...
-	NSString *commitShortDescription = [aNode localizedShortDescription];
-	if (commitShortDescription == nil)
-		commitShortDescription = @"";
-	
-	NSMutableDictionary *md = [@{kCOCommitMetadataShortDescriptionArguments : @[commitShortDescription]} mutableCopy];
+	NSMutableDictionary *md = [NSMutableDictionary new];
+	NSString *identifier = [(COCommandGroup *)aNode metadata][kCOCommitMetadataIdentifier];
+	NSString *shortDescription = [aNode localizedShortDescription];
+	NSNumber *inversedValue = aNode.metadata[kCOCommitMetadataNodeInversed];
+	BOOL inversed = inversedValue == nil || !inversedValue.boolValue;
```
ericwa: change to `BOOL inversed  = YES;` if we follow the idea in my other comment

ericwa:
This looks good, thanks.

Utilities/COCommitDescriptor.h
qmathe: ok. I applied these changes.

Undo/COUndoTrack.m
qmathe: I changed the name of the metadata key to kCOCommitMetadataUndoInitialBaseInversed.

Having just 'BOOL inversed = inverse' seems useless, since you can infer it easily by checking the kCOCommitMetadataUndoType (when the commit identifier is suffixed by 'undo', inversed is YES)

The inversed value I'm proposing could be computed by the client code, but it's easy to get it wrong and it would force you to access other commands than the current one, since you need to traverse the undo base command chain until you reach the initial base.

I use it to decide whether I applied a strikethrough style to the text. For example, suppose we have:

- Rename
- Selective Undo (Rename)
- Selective Undo (Selective Undo (Rename))

With this dedicated key, it can easily be presented as:

- Rename (kCOCommitMetadataUndoInitialBaseInversed == NO)
- ~~Rename~~ (kCOCommitMetadataUndoInitialBaseInversed == YES)
- Rename (kCOCommitMetadataUndoInitialBaseInversed == NO)

Utilities/COCommitDescriptor.h
qmathe: Done.

ericwa:
I think this is ready to merge. I thought maybe Typewriter would need some updates but it doesn't seem like it.

BTW, now the selective-undo/redo, and undo/redo commit descriptors here are deprecated: https://github.com/etoile/CoreObject/blob/master/English.lproj/Commits/org.etoile.CoreObject.json#L9

qmathe:
These changes shouldn't mess up past history or require new changes, they just give the ability to override the default presentation by writing some dedicated code. The old selective-undo/redo and undo/redo commit descriptors are still used by default, see -[COCommandGroup localizedShortDescription]. We now interpolate the short description for undo/redo operations at the time we access it, rather than at commit time.

For Typewriter, selective undos (done while using this CoreObject branch) could have broken short description when inspecting a branch history. I think -[CORevision localizedShortDescription] needs to be updated in the same way I did for -[COCommandGroup localizedShortDescription]. See the lines:

if (operationIdentifier != nil)
{
    COCommitDescriptor *operationDescriptor =
      [COCommitDescriptor registeredDescriptorForIdentifier: operationIdentifier];

ericwa:
>  The old selective-undo/redo and undo/redo commit descriptors are still used by default

That's not quite accurate, at least for writing, is it? If -[track undo] needs to synthesize a commit, it now uses the commit descriptor identifier of the node being undone, instead of "org.etoile.CoreObject.undo". Also, -[track undoNode:] no longer uses "org.etoile.CoreObject.selective-undo", same for redo.

Did you consider writing the identifiers ("org.etoile.CoreObject.undo/redo/selective-undo/selective-undo" the same as before, deleting kCOCommitMetadataUndoType, but putting the identifier of the undone/redone node in a separate key like "kCOCommitMetadataUndoOriginalIdentifier"?

I don't have much of a preference, although the scheme I describe in the last paragraph is closer to what we had before. Maybe what you have now is a bit cleaner, not sure.

qmathe:
Yes, we don't use old selective-undo/redo and undo/redo descriptors at commit time anymore. 'deprecated' gave me the impression you thought they were not used at all anymore, so my remark was more about the fact we still depend on them being present for synthesizing short descriptions lazily.

We could use kCOCommitMetadataUndoOriginalIdentifier. We would just need to adjust COCommandGroup code that generates the short description. 
I didn't take this approach to maintain  relationship between kCOCommitMetadataIdentifier  and kCOCommitMetadataShortDescription/kCOCommitMetadataShortDescription. With this identifier, you can retrieve the descriptor to interpret the short description and its arguments. It seemed these three metadata are at the same semantic level, while the undo/redo type/identifier expresses something a "higher level". However all that is pretty subtle… ;-) so I don't mind switching to kCOCommitMetadataUndoOriginalIdentifier.

ericwa:
My only slight issue with the current design is, if you think of kCOCommitMetadataIdentifier as the top-level key that defines what the commit does, then for undo commits this will be wrong - what the commit does will be the inverse of what kCOCommitMetadataIdentifier suggests. I agree about how the old undo/redo commit descriptors with kCOCommitMetadataIdentifier is mixing semantic levels. Anyway, in either case you need to read more than just the kCOCommitMetadataIdentifier key to discover what the commit is doing, so the difference between the two designs is pretty minimal. I'm fine with either one.

> One thing I forgot to mention is that… Keeping command groups as self-contained objects that can be presented without looking up other command groups, seems important to limit the complexity in history compaction code. That's why I took the decision to duplicate metadata such as kCOCommitMetadataShortDescription/kCOCommitMetadataShortDescription for selective undo/redo commits.

I agree, it's also important to keep them self contained because the other undo nodes could be deleted eventually.

qmathe:
I merged the changes. I kept the commit metadata keys as they were, but I fixed -[CORevision localizedShortDescription] to return a valid description for selective/undo operations.

Thanks for the review :) Let me know if you encounter any issues.

PR #29, opened 2015-11-21
COObjectGraphContext: properly discard objects when deallocated.

ericwa:
Replace any dangling references in other object graph contexts to the one being deallocated with [COPath brokenPath] placeholder objects.

### DIFF
--- a/Core/COObject+RelationshipCache.m
+++ b/Core/COObject+RelationshipCache.m
@@ -61,6 +61,12 @@ static BOOL isPersistentCoreObjectReferencePropertyDescription(ETPropertyDescrip
 		}
 		else
 		{
+            if (![obj isKindOfClass: [COObject class]])
+            {
+                NSLog(@"%@: Warning, ignoring non-COObject instance %@ in %@ of %@", NSStringFromSelector(_cmd), obj, aProperty.name, self);
+                return;
+            }
+            
 			[[obj incomingRelationshipCache] removeReferencesForPropertyInSource: [aProperty name]
 																	sourceObject: self];
 		}
--- a/Core/COObject+RelationshipCache.m
+++ b/Core/COObject+RelationshipCache.m
@@ -63,6 +63,12 @@ static BOOL isPersistentCoreObjectReferencePropertyDescription(ETPropertyDescrip
 		{
             if (![obj isKindOfClass: [COObject class]])
             {
+                // After inserting an illegal type of object into a collection that fails validation
+                // (throwing an exception), we don't currently remove the invalid object.
+                // (see -testNullDisallowedInCollection in the test suite.)
+                // This is a hack so that -[COObjectGraphContext dealloc] can complete
+                // without throwing an exception below, because the invalid object doesn't respond
+                // to -incomingRelationshipCache.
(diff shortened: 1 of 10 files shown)


ericwa:
Hmmm I'm not sure why the travis test runner is failing, all tests actually pass on my laptop and in the travis log.

Core/COObject+RelationshipCache.m
```diff
@@ -61,6 +61,12 @@ - (void) removeCachedOutgoingRelationshipsForCollectionValue: (id)obj
 		}
 		else
 		{
+            if (![obj isKindOfClass: [COObject class]])
```
qmathe: Is this for NSNull or COBrokenPath? Could to be nice to have a comment that gives an example or may be expand the log description.

Tests/Core/TestObjectGraphContext.m
```diff
@@ -660,4 +675,94 @@ - (void) testCrossContextReferenceSerializedAsNull
...
+    // add a child in ctx1
+    OutlineItem *ctx1obj = [[OutlineItem alloc] initWithObjectGraphContext: ctx1];
+    ctx1obj.label = @"ctx1obj";
+    
+    // create a ctx2
+    
+    COObjectGraphContext *ctx2 = [COObjectGraphContext new];
+    OutlineItem *ctx2root = [[OutlineItem alloc] initWithObjectGraphContext: ctx2];
+    ctx2root.label = @"ctx2root";
+    [ctx2 setRootObject: ctx2root];
```
qmathe: May be it's possible to replace -setRootObject: here and elsewhere in the method with .rootObject =

iirc I don't think the compiler would ask for an upcast, but I could be wrong.

qmathe:
All these changes look really good! 

Here is the test suite result on my computer btw:

Took 31884 ms
2015-11-25 01:59:05.883 Test[31368:774413] Result: 91 classes, 749 methods, 5329 tests, 0 failed, 0 exceptions

My only complaint would be that the indenting of the new code is bit broken Looks like the indenting is done with spaces instead of tabs. It's visible in the Github diff view or when changing the tabsize to something else than 4 characters.

Feel free to merge it and thanks again for sorting out this stuff :)

Core/COObject+RelationshipCache.m
ericwa: I added a comment to explain the purpose of this hack..

This case only occurs during the test suite for NSNull, when we explicitly insert it in the -testNullDisallowedInCollection to check that an exception is thrown. There's nothing special about NSNull though, it could be any ObjC class that we don't support adding to a particular collection.

The problem is, even though an exception is thrown at insertion time, the illegal object still gets inserted. Now that we clean up properly when an object graph is dealloced, this method was running and throwing another exception, when it tried to call `-incomingRelationshipCache` on the collection element. In fact, this issue was breaking UnitKit, because the exception was being thrown when an autorelease pool was drained that wasn't protected by a `@try`.

qmathe:
It's a bit crazy to see that removing 'Warning:' solved the travis build issue :) Perhaps it was getting confused by a double semi colon in output, when combining selector name + warning. For example: -addObject::Warning

PR #17, opened 2015-03-31
Add a simple binary format for serializing a COItemGraph, exposed via fu...

ericwa:
...nctions:

NSData *COItemGraphToBinaryData(id<COItemGraph> aGraph);
COItemGraph *COItemGraphFromBinaryData(NSData *binarydata);

Probably could use more error checking

PR #31, opened 2016-01-19
History compaction placeholder node

qmathe:
Here is a quick summary of the changes I did to improve history compaction:

- made possible to compact all commands so the only remaining node is the placeholder node
- fixed compaction to behave in a predictable way with a pattern undo track
  - child track current nodes can now be discarded (thanks to the first change)
- improved public COUndoTrackHistoryCompaction API and simplified implementation
- improved COUndoTrackStore to automatically finalize empty undo tracks rather than keeping them around
- fixed crash when attempting to reload a finalized persistent root (due to an outdated distributed notification)

Concerning this last change (see COPersistentRoot.m diff), I stumbled on it while testing the backup in Placeboard. It was a tricky bug to track down happening 1 time out of 50. When a backup is loaded in Placeboard, the entire model infrastructure operating in background (model factory, synchronization, compaction) of the app is recreated from scratch. This means every COEditingContext, COSQLiteStore and COUndoTrackStore instances are discarded at this time. 

However it seems ARC turns autorelease calls into normal release calls very often (49 time out of 50). So sometimes old COEditingContext will linger around a bit longer and they will catch distributed notifications posted by the new COSQLiteStore. Since old and new stores both have the same UUID (since they share the same store content but at different point in time), the distributed notification is not ignored. If the old store contains a deleted persistent root in memory (but previously finalized on disk) and the new store contains the same persistent root alive, when a compaction is run against the new store, the old store will catch the notification that reports this persistent root as compacted and attempt to reload it with:

	for (ETUUID *persistentRootUUID in compactedPersistentRootUUIDs)
	{
		COPersistentRoot *loaded = [_loadedPersistentRoots objectForKey: persistentRootUUID];

		[loaded storePersistentRootDidChange: notif isDistributed: isDistributed];
		hadChanges = YES;
	}

If loaded corresponds to a finalized to a persistent root in the old store, a crash waiting us ahead. The fix I committed is probably not the most correct one. A better one is probably to uncomment this TODO:

	for (ETUUID *persistentRootUUID in finalizedPersistentRootUUIDs)
	{
		__unused COPersistentRoot *loaded = [_loadedPersistentRoots objectForKey: persistentRootUUID];

		// TODO: [self unloadPersistentRoot: loaded]
		hadChanges = YES;
	}

However I'm not sure we cannot have other issues triggered by such outdated distributed notifications and mismatched store states between memory and disk. What do you think?

I considered adding stuff like a token marking every new store instances at initialization and checking it in the editing context, but this wouldn't work well when we have two different store objects pointing to same store file on disk. (either in the same application or in two different ones).

I guess the best solution is to make the notification handler solid enough to survive any out-of-order delivery of notifications. 

Another solution is to make an explicit -invalidate call that discards the observers when you are done with an editing context.

If you want to play with this issue, comment out the changes in COPersistentRoot.m and an exception should occur right after running -testCompactionsInterleavedWithCommitsAndUndoRedo:

#3	0x000000010037a00b in -[COSQLiteStore itemGraphForRevisionUUID:persistentRoot:] at /Users/qmathe/reps/Etoile/Frameworks/CoreObject/Store/COSQLiteStore.m:559
#4	0x000000010032c381 in -[COPersistentRoot setCurrentBranchObjectGraphToRevisionUUID:persistentRootUUID:] at /Users/qmathe/reps/Etoile/Frameworks/CoreObject/Core/COPersistentRoot.m:211
#5	0x000000010032c4c6 in -[COPersistentRoot reloadCurrentBranchObjectGraph] at /Users/qmathe/reps/Etoile/Frameworks/CoreObject/Core/COPersistentRoot.m:219
#6	0x000000010033335f in -[COPersistentRoot storePersistentRootDidChange:isDistributed:] at /Users/qmathe/reps/Etoile/Frameworks/CoreObject/Core/COPersistentRoot.m:878
#7	0x0000000100319b47 in -[COEditingContext storePersistentRootsDidChange:isDistributed:] at /Users/qmathe/reps/Etoile/Frameworks/CoreObject/Core/COEditingContext.m:914
#8	0x0000000100318e2f in -[COEditingContext distributedStorePersistentRootsDidChange:] at /Users/qmathe/reps/Etoile/Frameworks/CoreObject/Core/COEditingContext.m:793
#9	0x00007fff8b47a45c in __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ ()
#10	0x00007fff8b437f8e in ____CFXNotificationPostToken_block_invoke ()
#11	0x00007fff8b3da8ec in __CFRUNLOOP_IS_CALLING_OUT_TO_A_BLOCK__ ()
#12	0x00007fff8b3cc9f5 in __CFRunLoopDoBlocks ()
#13	0x00007fff8b3cc536 in __CFRunLoopRun ()
#14	0x00007fff8b3cbbd8 in CFRunLoopRunSpecific ()
#15	0x00007fff8ad78b29 in -[NSRunLoop(NSRunLoop) runMode:beforeDate:] ()
#16	0x00007fff8ad96d9e in -[NSRunLoop(NSRunLoop) runUntilDate:] ()
#17	0x00000001005054b8 in -[UKRunner runTest:onObject:class:] at /Users/qmathe/reps/Etoile/Frameworks/UnitKit/FrameworkSource/UKRunner.m:332

### DIFF
--- a/Core/COPersistentRoot.m
+++ b/Core/COPersistentRoot.m
@@ -849,6 +849,13 @@ cheapCopyPersistentRootUUID: (ETUUID *)cheapCopyPersistentRootID
     
     COPersistentRootInfo *info =
 		[[self store] persistentRootInfoForUUID: [self UUID]];
+	
+	/* If we are receiving a changed/compacted notification but the persistent
+	   root has been finalized in the meantime (distributed notifications are 
+	   delivered in LIFO order). */
+	if (info == nil)
+		return;
+
     _savedState = info;
     
     for (ETUUID *uuid in [info branchUUIDs])
--- a/Store/COHistoryCompaction.h
+++ b/Store/COHistoryCompaction.h
@@ -68,6 +68,9 @@
  *
  * Branches not previously marked as deleted are ignored.
  *
+ * To finalize a parent branch, the child branches must all appear in the
+ * returned set and have been previously marked as deleted.
+ *
  * To attempt finalizing all branches, return -compactableBranchUUIDs.
  */
 @property (nonatomic, readonly) NSSet *finalizableBranchUUIDs;
@@ -87,7 +90,19 @@
 /** @taskunit Revision Status */
 
 
-- (NSSet *)deadRevisionUUIDsForPersistentRootUUIDs: (NSArray *)persistentRootUUIDs;
+/**
+ * Returns the live revision sets per persistent root.
+ *
+ * When compacting the store, all revisions older than the oldest revision in 
+ * this set will be discarded, and all revisions newer will be kept.
+ *
+ * Any branch with revisions appearing in this set will be kept.
+ *
+ * For a branch forked from another branch, the revision corresponding to the 
+ * branch creation is always treated as a live one until the forked branch is 
+ * deleted (not yet implemented). No matter which revisions you return, this 
+ * ensures you cannot accidentally create detached branches.
+ */
 - (NSSet *)liveRevisionUUIDsForPersistentRootUUIDs: (NSArray *)persistentRootUUIDs;
 
 
--- a/Tests/Undo/TestUndoTrackHistoryCompaction.m
+++ b/Tests/Undo/TestUndoTrackHistoryCompaction.m
@@ -212,7 +212,7 @@
 	                             expectingCompaction: compaction];
 
 	UKObjectsEqual(liveRevs, newRevs[persistentRoot.UUID]);
-	UKObjectsEqual(liveCommands, track.nodes);
+	UKObjectsEqual(liveCommands, [track.nodes subarrayFromIndex: 1]);
 	UKObjectsEqual(liveCommands, track.allCommands);
 	
 	[self checkUndoRedo];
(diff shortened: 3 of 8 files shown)


ericwa:
On the issue where the autoreleased editing context is receiving unwanted notifications, this is sort of similar to the problem of objects in a garbage-collected languages holding resources for indeterminate amounts of time.

At first glance, the cleanest solution is just add a -close method to COEditingContext, this would close the database (probably just do `_store = nil;` which should handle closing the DB?).

qmathe:
-close or -invalidate would work just fine. The issue is not really about releasing the store, but about stopping observing the store with NSDistributedNotificationCenter, so the editing context doesn't receive notifications matching the store UUID.

Setting the store to nil is probably a safe move, although I don't expect keeping the store object around to be an issue, since the store doesn't observe any notifications/events and I don't remember it to have methods with a completion block… in the sense that operations executed on the DB queue are always initiated with dispatch_sync(), which makes the main que waits and prevents the store to be released while executing a background operation (we don't allow allow releasing the store outside of the main thread afaik).

ericwa:
I think this is fine to merge. I don't know the undo track garbage collection too well, but it sounds good.
One of us can do the -close/-invalidate in a separate PR

qmathe:
Thanks for the merge.

-close/-invalidate is 5 five lines of code to add, so we can probably commit the change directly into master.

PR #32, opened 2016-01-19
COObject+Accessors rewrite

ericwa:
See commit comments

### DIFF
--- a/Core/COObject+Accessors.m
+++ b/Core/COObject+Accessors.m
@@ -51,18 +51,18 @@ static void genericSetter(id self, SEL theCmd, id value)
 {
     // FIXME: Same comment as the genericGetter
 
-    const char *setter_cstring = sel_getName(theCmd);
-    size_t setter_cstring_len = strlen(setter_cstring);
-    char key_cstring[setter_cstring_len];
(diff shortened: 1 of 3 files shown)


qmathe:
The code looks good, feel free to merge it. Thanks a lot for working on that :)

ericwa:
thanks, I think I fixed all of the style problems.

qmathe:
I just took a look at the performance improvements.

At launch time, the time spent in +resolveInstanceMethod: went down from 200/300 ms to 1ms! 

The performance improvement is also visible elsewhere sometimes. For example, when navigating accross a big history, this change can cut 50ms too.

Beside PropertyToSetter(), the two following lines were also important bottlenecks:

NSString *attributes = [NSString stringWithUTF8String: property_getAttributes(property)];
BOOL isDynamic = ([attributes rangeOfString: @"D"].location != NSNotFound);

ericwa:
Nice. The `+resolveInstanceMethod:` changes should only affect startup time (or the first time classes are used).

The change in `genericSetter` will make setters faster since there's no longer any NSString manipulation, and no use of format strings, just creating an NSString from a C string.

PR #35, opened 2016-01-24
COPersistentRoot: remove _lastTransactionID ivar and store this in a dictionary in COEditingContext.

ericwa:
This is so we can still retrieve the last transaction ID even if the persistent root is deallocated.


qmathe:
Everything looks ok.

ericwa:
Cool, I fixed the order of CORegisterCoreObjectMetamodel as you suggested

PR #36, opened 2016-01-27
History navigation optimization

qmathe:
This speeds up history navigation by 2x when inner objects have many properties, and also when object graphs contains several inner objects (rather than a single one used as a root object).

There is a new benchmark method to check the optimizations: -testGoToOldestAndNewestNodesInHistoryWithManyPropertiesPerEntity

Some optimizations are based on recent changes I committed to EtoileFoundation.

I could probably optimize it further, but this would require more substantial changes and complex optimizations. So before that, I'd like to remove all EtoileUI-compatibility code that currently slows down the result a bit (e.g. -serializationSetterForProperty: or -isCoreObject:), and also see the result of incremental loading.

### DIFF
--- a/Core/COObject.m
+++ b/Core/COObject.m
@@ -1202,22 +1202,6 @@ See +[NSObject typePrefix]. */
 	[self didChangeValueForKey: property atIndexes: indexes withObjects: objects mutationKind: mutationKind];
 }
 
-// TODO: Move this method to ETModelDescriptionRepository
-- (ETEntityDescription *) entityDescriptionForObject: (id)anObject
-{
-	if ([anObject isKindOfClass: [COObject class]])
-	{
-		// special case to support the case when we're using COObject class
-		// and not a user-supplied subclass
-		return [(COObject *)anObject entityDescription];
-	}
-	else
-	{
-		return [[_objectGraphContext modelDescriptionRepository]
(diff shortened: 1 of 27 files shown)


ericwa:
Great, thanks for working on this. I'll take a closer look.
Looks like there is a assertion failing in BenchmarkCoreObject, only when this branch is merged with `master`.

Core/COObject.m
```diff
@@ -868,15 +902,19 @@ - (id)valueForVariableStorageKey: (NSString *)key
 	id value = [_variableStorage objectForKey: key];
 	
 	// Convert value stored in variable storage to a form we can return to the user
-	if (value == [NSNull null])
+	if (value == nil)
+	{
+		return aNotFoundMarker;
+	}
+	else if (value == null)
```
ericwa: This is a bit confusing at first glance, maybe rename the cached NSNull instance from `null` to `static_nsnull` or something?

ericwa:
Looks good. I'm going to take a look at the assertion failure in Benchmark now

ericwa:
OK I just pushed 3 commits: merging in master, fixing the failure in Benchmark, and renaming null to `cachedNSNull`. Hope this doesn't cause any inconvenience.

qmathe:
Thanks for the merge and tweaks, Eric. Everything is ok for me.

PR #37, opened 2016-01-27
COPrimitiveCollection: rewrite `mutable` property.

ericwa:
- now there are a pair of -beginTemporaryModification and -endTemporaryModification calls that support nesting properly.
- -mutableCopy returns an instance that is always mutable.

This is something I'll need for the persistent root lazy loading, but it's also cleaner IMO.


qmathe:
Looks good, feel free to merge it.

PR #39, opened 2016-01-27
Persistent root lazy loading

ericwa:


ericwa:
#### Backward/inverse cross-persistent root references 

I saw in TODO.md we anticipated one of the problems I ran into:
>- Persistent root faulting; currently the entire store is loaded in memory
>
>	- A challenge will be supporting cross-persistent root reference inverses. i.e. suppose we open a store, load a document persistent root into memory. Accessing the derived property "document.tags" requires a search query over the current revisions of the current branches of all persistent roots in the whole store (at least conceptually). The store keeps an index of cross-persistent references anticipating this problem, but it's not currently used. The most straightforawrd solution would be to load every persistent root that has ever had a cross-reference to document at the same time as when we load "document".

Looking at it now, it seems messy to try to keep track of the backward references in the store. 

What happens now is the backward references like "document.tags" will be empty if the tag persistent root is not loaded, see -testSourcePersistentRootLazyLoading in TestUnivaluedRelationshipWithOpposite.m: https://github.com/etoile/CoreObject/pull/39/files#diff-17ef14aa07843c2212fc199cd348410eR458 

This does leak information about which persistent roots are faulted and which aren't.
In other words, properties like "document.tags" are now fuzzy (since they might be missing some objects belonging to unloaded persistent roots) rather than exact as they were previously.

Is this an acceptable limitation? I'm not sure.. it could be unacceptable for a property "document.tags" since the only way to get accurate info is to load all of the tag persistent roots. On the other hand, we've previously discussed how using that model for tagging is a poor choice (storing the list of tagged documents in the tags).

ericwa:
Another tricky aspect of this patch is the question of when to trigger the unfaulting.

Currently:
- The first call to `-persistentRoots` on a COEditingContext loads all persistent roots in the store at that time. See `loadAllPersistentRootsIfNeeded`. This is implemented by an ivar. Note that it only checks the store once, future calls will never pull the list of persistent roots from the store. We rely on notifications for future updates from other contexts.
- Accessing a COObject property will trigger lazy loading of any persistent roots referenced by that property. This is implemented in `valueForVariableStorageKey:notFoundMarker:shouldLoad:`, by calling `replaceReferencesWithLoadedObjectsForKey:collection`.  Currently, if there are unloaded references (COPath), we re-try every time the property is accessed. This should be fixed I guess, otherwise it may cause a slowdown (but only in some scenarios).

ericwa:
I'm pretty happy with this now. The implementation is a bit ugly, but we want this feature sooner or later, so I think it's better to do refactoring later.

qmathe:
ok, cool ! I'll take a look at it tomorrow, time to sleep a bit ;)

Core/COEditingContext.m
```diff
@@ -911,26 +964,9 @@ a correct deletion status (critical to update the cross
 		}
 		else if (![deletedPersistentRootUUIDs containsObject: persistentRootUUID])
 		{
-			/* For a finalized persistent root, insertedOrUndeleted is nil */
-			COPersistentRoot *insertedOrUndeleted = [self persistentRootForUUID: persistentRootUUID];
-
-			if (insertedOrUndeleted != nil)
```
qmathe: If this piece of code is removed, how do you handle updating references to persistent root deleted done in another editing context?

Benchmark/TestHistoryNavigationPerformance.m
```diff
@@ -166,13 +166,13 @@ - (void)testGoToFirstCommitNodeToReloadManyDeletedPersistentRoots
...
 	
-	UKIntsEqual(BIG_NUM_PERSISTENT_ROOTS, ctx2.loadedPersistentRoots.count);
+	UKTrue(ctx2.loadedPersistentRoots.isEmpty);
 	
 	NSTimeInterval goToNodeTime = [[NSDate date] timeIntervalSinceDate: startDate];
 	NSLog(@"Time to go to first commit node on undo track: %0.2fs", goToNodeTime);
 	startDate = [NSDate date];
 	
-	UKTrue(goToNodeTime < 3.0); // FIXME: 1.0
+	UKTrue(goToNodeTime < 0.5);
```
qmathe: This is nice :)

qmathe:
Looks pretty neat and not too complex, thanks for tackling this :)

I just made some small comments. I haven't read the tests carefully.

The backward/inverse cross-persistent root references looks like a pretty big issue, it's going to break stuff in Placeboard :-/ So if you don't mind I'd like to push back merging this pull request, until I'm done with the first release and I get more time to investigate this issue and what we could do about it.

For the lazy loading, I was envisioning a more fine-grained loading than unfaulting all references the first time a collection is accessed. I mean something like checking the object type in -[COMutableArray objectAtIndex:] and similar methods of other COPrimitiveCollections, in order to call -crossPersistentRootReferenceWithPath: and replace if needed the COPath instance by a live object in the collection. 
This way, you could have a big list and the references would get unfaulted as you scroll through the list in the UI. You can see my reply to 'CoreObject performance issues with big histories' on 24/01 where I explain it a bit more in term of implementation. The problem with what I suggested is to ensure performance doesn't suffer when you want to load the all objects in a single pass (as done currently).
What do you think?

Core/COEditingContext.m
ericwa: That branch of the if() only executes if the persistent root is not currently loaded, so I guess at that point any references to it must have already been fixed

ericwa:
>The backward/inverse cross-persistent root references looks like a pretty big issue, it's going to break stuff in Placeboard :-/ So if you don't mind I'd like to push back merging this pull request, until I'm done with the first release and I get more time to investigate this issue and what we could do about it.

The only fix is to keep a table of cross references in the store, otherwise you would have to load all persistent roots to find cross references. Then when you access a derived property like -parentCollections, it would go to a special unfaulting code path that checks the table for persistent roots with possible references, and then unfaults all of those. 

The table should contain two columns, cross-reference target persistent root (uuid), cross-reference source backing store uuid. 

>For the lazy loading, I was envisioning a more fine-grained loading than unfaulting all references the first time a collection is accessed. I mean something like checking the object type in -[COMutableArray objectAtIndex:] and similar methods of other COPrimitiveCollections, in order to call -crossPersistentRootReferenceWithPath: and replace if needed the COPath instance by a live object in the collection. 

The only way I see it working is to have another type of fault marker. It would store the same data as COPath, but would count towards the collection length when -count is called. At first I discarded the idea because it seemed too complex, but maybe it's not too bad.

Core/COEditingContext.m
```diff
@@ -369,6 +392,11 @@ - (void)undeletePersistentRoot: (COPersistentRoot *)aPersistentRoot
  */
 - (void)updateDeadRelationshipCacheForUndeletedPersistentRoot: (COPersistentRoot *)aPersistentRoot
 {
+	// FIXME:
```
qmathe: I think this code should be moved to -updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted: and rewritten a bit since including the deleted branches as I did it seems wrong. For this last point, I could be overlooking something. 

-updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted:
{	
	// more code before

	if (isDeleted)
		return;

	for (COBranch *branch in aPersistentRoot.branches)
	{
		[_deadRelationshipCache removePath: [COPath pathWithPersistentRoot: aPersistentRoot.UUID
		                                                            branch: branch.UUID]];
	}
	[_deadRelationshipCache removePath: [COPath pathWithPersistentRoot: aPersistentRoot.UUID]];
}

Is this correct or am I missing something?

ericwa: I'm not sure, sorry :-(

I am pretty sure I was forced to add that `#If 0` to make some tests pass, so if you re-enable this code you should see some failure hopefully showing why I had to disable it.

If it works to move that code to `-updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted: ` feel free.

qmathe: It makes sense to comment it out with #if 0, I would expect -updateDeadRelationshipCacheForUndeletedPersistentRoot: to break some tests.

I'll try to move the code to -updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted: and see how it goes.

qmathe: I ended up removing this code entirely, since it's not needed at all. I wrote some tests to ensure the dead relationship cache is still properly cleaned. See my last commit in this pull request.

Core/COEditingContext.m
```diff
@@ -412,8 +416,23 @@ - (void)updateCrossPersistentRootReferencesToPersistentRoot: (COPersistentRoot *
...
-		|| (!aBranch.isCurrentBranch && aBranch.persistentRoot == aPersistentRoot));
+	// NOTE: This is used by lazy loading now, and the following assertion no longer holds.
+	// Is this a problem?
+	//NSParameterAssert(aBranch == nil
+	//	|| (!aBranch.isCurrentBranch && aBranch.persistentRoot == aPersistentRoot));
+	
+	// See documentation above
+	if (isDeletion)
+	{
+		// TODO: For an uncommitted persistent root and branch, could be better if
```
qmathe: For this TODO, we can rewrite -deletePersistentRoot like that:

```
- (void)deletePersistentRoot: (COPersistentRoot *)aPersistentRoot
{
    if ([_persistentRootsPendingUndeletion containsObject: aPersistentRoot])
    {
        [_persistentRootsPendingUndeletion removeObject: aPersistentRoot];
    }
    else
    {
        // NOTE: Deleted persistent roots are removed from the cache on commit.
        [_persistentRootsPendingDeletion addObject: aPersistentRoot];
    }

	[self updateCrossPersistentRootReferencesToPersistentRoot: aPersistentRoot
	                                                   branch: nil
	                                                isDeleted: YES];
	
	if ([aPersistentRoot isPersistentRootUncommitted])
    {
		[_persistentRootsPendingDeletion removeObject: aPersistentRoot];
        [self unloadPersistentRoot: aPersistentRoot];
    }
}
```

All the tests still pass when I apply this change. We could do a similar change to -deleteBranch:. What do you think?

qmathe: Another thing that suprised me is that no tests were broken, even though isDeletion was sometimes incorrect. I corrected the code in the callers, but I haven't investigated why it was working when I wouldn't have expected it.

Core/COEditingContext.m
```diff
@@ -947,6 +964,9 @@ cache in the same way on insertion and undeletion (insertion
 		__unused COPersistentRoot *loaded = [_loadedPersistentRoots objectForKey: persistentRootUUID];
 
 		// TODO: [self unloadPersistentRoot: loaded]
+		// Temporary hack until we have -unloadPersistentRoot:
+		[_lastTransactionIDForPersistentRootUUID removeObjectForKey: persistentRootUUID];
```
qmathe: May be we should implement -unloadPersistentRoot if it's not too difficult?

Core/COEditingContext.m
qmathe: I figured out why the tests were passing, even when isDeletion was incorrect. I found a HACK in -replaceReferencesToObjectIdenticalTo:withObject: which can now be removed without breaking the test suite.

```
		// HACK: Now using -replaceReferencesToObjectIdenticalTo in the lazy loadying system
		// rather than just deletion/undeletion, so we need the logic in -crossPersistentRootReferenceWithPath
		// that checks whether branches/persistent roots are deleted
		COObject *testReplacement = [self.editingContext crossPersistentRootReferenceWithPath: object shouldLoad: NO];
		if (testReplacement == nil)
		{
			return;
		}
```

ericwa:
Weird, I can't reproduce those test failures locally..

ericwa:
OK, with any luck  	6b53f7e will fix the test failures on travis.

ericwa:
OK, good, it's fixed. apparently `aDict[key] = nil;` doesn't work on OS X 10.9 which is what our travis builds currently run.

I can't find documentation on that off hand.. although we could use a newer OS X on travis if you want: https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version

Core/COEditingContext.m
ericwa: I'd favour doing that in another PR, since this one is pretty complex already.
But feel free to add it to this PR if you feel like implementing it now.

Core/COEditingContext.m
ericwa: Cool.. the hack was necessary until your commit bec8cb9c2e89d850f363fb88fb0cd2d3de915bc6
(-testTargetPersistentRootDeletion at TestOrderedRelationship.m:268 would fail without the hack).

qmathe:
Thanks for the fix. Looks like Apple changed -[NSMutableDictionary setObject:forKeyedSubscript:] behavior in recent OS X versions: http://stackoverflow.com/a/32679535

Using -removeObject:forKey: sounds good.

Core/COEditingContext.m
qmathe: Agreed, it's probably better to do it in another PR.

ericwa:
I'm fine with merging this, anything else that needs to be looked at?

qmathe:
I just made some extra cleaning, everything looks ok to me.

PR #46, opened 2016-04-30
Custom migration support

qmathe:
The existing schema migration API works well 95% of the time when changing the schema, but in some special cases it's not enough and the item graph needs to be patched directly (e.g. to change items entities/properties accross packages without altering the metamodel). 

To do this, I have exposed COSchemaMigrationDriver API and made possible to initialize COObjectGraphContext with a custom driver subclass. A driver subclass can override `-migrateItems:` to implement a custom migrations strategy. Here is an example in Swift:

```
class PlaceboardMigrationDriver : COSchemaMigrationDriver {

	// For Placeboard 1.0.2 -> 1.0.3
	//
	// Forgot to subclass COTagLibrary entity located in CoreObject package when creating Placeboard metamodel, this migration fixes the item graph directly.
	func migrateTagLibrary(storeItems: [AnyObject]!) -> [AnyObject]! {
		
		// The guard clause below is equivalent to:
		// if storeItems.count != 1 || item.packageVersion != 0 || item.entityName != "COTagLibrary" { return storeItems }
		guard let item = storeItems.first
		    where storeItems.count == 1 && item.packageVersion == 0 && item.entityName == "COTagLibrary" else {
				return storeItems
		}
		precondition(item.packageName == "org.etoile-project.CoreObject")

		let newItem = item.mutableCopy() as! COMutableItem
		let package = modelDescriptionRepository.descriptionForName(NSBundle.mainBundle().bundleIdentifier) as! ETPackageDescription
			
		newItem.entityName = "PWTagLibrary"
		newItem.packageName = package.name
		newItem.packageVersion = Int64(package.version)
			
		return [newItem]
	}

	override func migrateItems(storeItems: [AnyObject]!) -> [AnyObject]! {
		let patchedItems = migrateTagLibrary(storeItems)

		return super.migrateItems(patchedItems)
	}
}
```

To create editing contexts that leverage the custom driver above:
```
		let ctx = COEditingContext(store: store,
		   modelDescriptionRepository: repo,
		         migrationDriverClass: PlaceboardMigrationDriver.self,
		               undoTrackStore: COUndoTrackStore(URL: undoTrackStoreURL))
```

### DIFF
--- a/Core/COEditingContext.h
+++ b/Core/COEditingContext.h
@@ -153,6 +153,7 @@ typedef NS_ENUM(NSUInteger, COEditingContextUnloadingBehavior) {
 	@private
 	COSQLiteStore *_store;
 	ETModelDescriptionRepository *_modelDescriptionRepository;
+	Class _migrationDriverClass;
 	/** Loaded (or inserted) persistent roots by UUID */
 	NSMutableDictionary *_loadedPersistentRoots;
 	COEditingContextUnloadingBehavior _unloadingBehavior;
@@ -194,8 +195,8 @@ typedef NS_ENUM(NSUInteger, COEditingContextUnloadingBehavior) {
 /**
  * <init />
  * Initializes a context which persists its content in the given store,
- * manages it using the metamodel provided by the model description repository, 
- * and whose undo tracks are backed by the last store argument.
+ * manages it using the metamodel provided by the model description repository 
+ * and migration driver, and whose undo tracks are backed by the last store argument.
  *
  * When COObject entity description doesn't appear in the repository, this 
  * initializer invokes +newEntityDescription on COObject and its subclasses, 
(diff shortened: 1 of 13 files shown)


ericwa:
Looks fine, though it seems like a majority of the diff is threading the migration driver through various initializers so it can get to COObjectGraphContext.

Maybe it'd be better if `COObjectGraphContext` got the migration driver through COEditingContext like this?
```
[self.editingContext.migrationDriver migrateItem: ...]
```

qmathe:
I agree it'd be better, but I don't think it's possible. Basically it's the same problem that the one we have with the model description repository. Since we support creating transient object graphs without any editing context, this means we must reference both the repository and driver class right from  COObjectGraphContext.

May be there is a better way to handle this, any ideas?

ericwa:
Ah right. This looks fine to merge then.

In the longer run, we could make a protocol that covers all the external functionality COObjectGraphContext/COObject need from COEditingContext.

```
@protocol COEditingContextProtocol
-modelDescriptionRepository
-migrationDriver
...
@end

@interface COObjectGraphContext
- (id<COEditingContextProtocol>)editingContextProtocolObject;
@end
```
`editingContextProtocolObject` would either return a real COEditingContext, or in the case of a standalone COObjectGraphContext, just a "holder" object that holds references to the desired model description repo, migration driver, etc.

qmathe:
ok, I just merged the changes.

The editing context protocol object sounds like a good idea. I'll mention it in the TODO.

PR #50, opened 2016-08-30
Persistent root unloading

qmathe:
I finished to implement the unloading support, this was quite straightforward to implement.

-[COEditingContext unloadPersistentRoot:] should probably be changed to take a set of persistent roots. This would match the unload notification that reports a set of unloaded persistent roots. What do you think?

For -updateCrossPersistentRootReferences in COEditingContext, it could be better to rename `faulting`as `dead`and `isTargetFaulting`as `isTargetDead` to get a more uniform terminology, since we already have variables mentionning dead vs live refs in this method. isTargetFaulting sounds also a bit weird to my ears.

Let me know if you have any questions or comments.

### DIFF
--- a/Core/COBranch.h
+++ b/Core/COBranch.h
@@ -304,7 +304,12 @@ extern NSString * const kCOBranchLabel;
  * Use it to cause a "checkpoint" revision to be written.
  */
 @property (nonatomic, assign) BOOL shouldMakeEmptyCommit;
-
+/**
+ * Returns whether the editing context has relinquished
+ * control over this object. If YES, this instance can no longer be used
+ * and calling any methods may throw an exception
+ */
+@property (nonatomic, readonly) BOOL isZombie;
 
 /** @taskunit Undo / Redo */
 
--- a/Core/COBranch.m
+++ b/Core/COBranch.m
@@ -113,6 +113,10 @@ parentRevisionForNewBranch: (ETUUID *)parentRevisionForNewBranch
 
 - (NSString *)description
 {
+	if ([self isZombie])
+	{
+		return @"<zombie branch>";
+	}
 	return [NSString stringWithFormat: @"<%@ %p - %@ (%@) - revision: %@>",
 		NSStringFromClass([self class]), self, _UUID, [self label], [[self currentRevision] UUID]];
 }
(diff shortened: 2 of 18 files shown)


ericwa:
I'll make a few changes if they sound OK to you:
- I think we need `isZombie` properties on `COPersistentRoot` and `COBranch`.
- `forces:` sounds better if changed to `force:`
- I don't like setters having potential side effects when setting their current value.
    maybe `-[COPersistentRoot setDeleted:flag]` should start with `if (flag == [aPeristentRoot isDeleted]) return;`?

- `testUndeleteUncommittedPersistentRoot` and `testUndeleteUncommittedBranch` don't do what they sound like, they should mark the persistent root / branch as deleted, then undelete it (which fails because the persistent root / branch is a zombie).

qmathe:
Thanks for renaming forces: to force: and adding -isZombie to COBranch and COPersistentRoot. Both changes sound good.

I agree about adding a guard clause to -[COBranch/COPersistent setDeleted:] that returns immediately.  The test behavior you suggest for -testUndeleteUncommittedPersistentRoot/Branch sounds better too.

To give some extra background… The goal of -testUndeleteUncommittedPersistentRoot/Branch was to ensure -setDeleted: is a proper no-op when the branch or persistent root haven't been committed yet. As a side effect, this was also checking that -updateCrossPersistentRootReferencesToPersistentRoot:branch:isDeleted: can be safely called twice with the same arguments without breaking anything. 

With your changes, we don't have to worry about that, so it's better.

ericwa:
@qmathe Do you mind checking my last 2 commits, I made it so unloading a uncommitted persistent root creates a zombie.

I think prior to those commits, such a persistent root was already essentially a zombie, (the editing context had no remaining pointers to the COPersistentRoot), so with my changes we just null out the _parentContext pointer on the persistent root.

I guess I'll merge this now :)

qmathe:
The last 2 commits look fine to me.

COBranch seems to be missing -makeZombie though, I see no code that sets _persistentRoot to nil. We probably also need to update existing branch deletion/undeletion tests to check COBranch.isZombie.

Another stylistic change that could be good is to update COObject to use the same exact method names when dealing with zombies. For example, to rename -markAsRemovedFromContext to -makeZombie, and -checkIsNotRemovedFromContext to -assertNotZombie (or -checkIsNotZombie).

We should probably change -checkIsNotRemovedFromContext to throw an exception, then document that any COObject methods can throw an exception when COObject.isZombie is YES. Just to be consistent with COPersistentRoot/COBranch.

PR #51, opened 2016-09-13
Added french localization for Commits

qmathe:
Switched to Base localization with English as development language

Fixed language directory lookup to be compatible with Base localization and preferred localizations


ericwa:
Hmm.. this test is failing for some reason?
```
- (void) testSelectiveRedoCommitDescriptor
{
	UKNotNil([COCommitDescriptor registeredDescriptorForIdentifier: @"org.etoile.CoreObject.selective-redo"]);
}
```

ericwa:
I think I fixed it

qmathe:
I committed the correct fix for the failing test in the master. 

I also added a little explanation that makes it clearer why the preferred localizations have to be requested to the main bundle.

Ideally we should have some tests to cover this, but this would require some refactoring to support a method like +reloadCommitDescriptorsForPreferredLocalizations:.

qmathe:
Thanks for the review btw :)

PR #52, opened 2016-09-13
Fixed replacement operation for sets and dictionaries for -commonWill…

qmathe:
…ChangeValueForProperty: fast path

Fixed -validateNewValue:propertyDescription: fast path when a serialization getter is used for a multivalued property

Documented -commonWillChangeValueForProperty: fast path

Note: with this change, keyed relationships should work, including bidirectional ones if we want to support them, COUnsafeRetainedDictionary is missing though.


Core/COObject.m
```diff
@@ -1305,6 +1323,9 @@ - (void)validateNewValue: (id)newValue
 	ETModelDescriptionRepository *repo = _objectGraphContext.modelDescriptionRepository;
 	BOOL isPersistentRelationship = propertyDesc.isPersistentRelationship;
 	BOOL isValidatableType = ([self serializationGetterForProperty: propertyDesc.name] == NULL);
+	
+	if (!isValidatableType)
+		return;
```
ericwa: Won't this skip a bunch of the ETAsserts below? maybe that could hide bugs in the future.

ericwa:
Feel free to merge.. if there is an easy way to avoid skipping the asserts I mentioned, that would be nice, otherwise probably not a huge issue.

Core/COObject.m
qmathe: It's more a temporary workaround until serialization accessors get removed. This hack is already present in the slow path.

This could hide bugs but basically serialization accessors can return arbitrary objects, and this validation code would need to be changed to check ETPropertyDescription.persistentType if we want to run all these assertions. Since the code is going away soon, I don't think it's worth the investment.

qmathe:
ok, thanks a lot for the review. I still need to write some tests to cover these changes though.

Core/COObject.m
ericwa: cool, sounds fine to me!

PR #57, opened 2016-09-22
Modern objc

qmathe:
Here is the result of migrating the code base to modern Objective-C syntax. The migration includes the following changes:

- objc literals like @[] and @{} and @(), except inside macros
- enforced public/package designated initializers with NS_DESIGNATED_INITIALIZER
- instancetype instead of id for initializers and other similar methods
- standardized property declarations everywhere as documented in HACKING
- property dot syntax
- !expression instead of expression == NO
- NSENUM/NSOPTIONS macros
- subscripting syntax when dealing with arrays and dictionaries

The framework code base has been mostly migrated with Xcode, while test and benchmark code, plus any code inside assertion macros have been migrated manually since the Xcode migration tool was unable to migrate them. Most of the commits I made are related to migrating the test code manually.

I doubt it's worth the pain to review entirely, since the diff is probably touching 30000 loc. However you might want to check the tweaks I made the HACKING document , in order to better document the property declaration rules, since coming up with sane rules is tricky.


ericwa:
Thanks! Looks good to merge.

Just noticed these suggestions in HACKING.md conflict with each other
> - readonly can be combined with copy or weak (see the rule before too)
> - don't declare a readonly property as copy (copy only describes if the 
object is copied by the setter)

Is it correct that when you implement the getter/setter yourself, the strong/weak/copy part of the property declaration is for documentation only?

qmathe:
Thanks for taking a look! I corrected the copy vs readonly issue in HACKING.

The only legitimate use case for (readonly, copy) is when overriding a superclass property declared as (readwrite, copy). If you omit _copy_ in the property declaration of the subclass, then you get a warning complaining about mismatched property attributes.

Overriding a (readonly) property with (readwrite, copy) won't cause any warnings. Same if you redeclare a (readonly) property as (readwrite, copy) in the implementation with a class extension.

qmathe:
When you implement the setter yourself, the copy part of the property declaration is just for documentation. For strong/weak, they apply to the ivar even when you implement the setter and getter. When you have no corresponding ivar, strong/weak are useless and imo that's why we shouldn't use them in these cases.

Note: for a weak property, iirc the corresponding explicit ivar must be marked as __weak otherwise the compiler will complain about a mismatch.

Issue #48, opened 2016-08-28
Support subscripting to read/write COItem attributes


qmathe:
Subscripting would be mapped to -valueForAttribute:, -setValue:forAttribute: and -removeValueForAttribute: (when assigning nil).

qmathe:
Closing this issue, since we decided against it in PR Item subscripting #72

PR #72, opened 2016-10-04
Item subscripting

qmathe:
I played a bit with adding subscripting support to COItem. Since the result doesn't look as good as I expected, I didn't continue my work in master, but created a PR.

Not sure it's worth to keep subscripting support for COItem, since it only help a bit in the tests… elsewhere keeping -valueForAttribute: often looks better since we use -typeForAttribute: at the same time. Also updating an attribute with subscripting is not a possibility, since we always call -setValue:forAttribute:ofType: and never just -setValue:forAttribute:.

I added subscripting support with 9dffe9f357189280fe24219074a5f4ba4aa7a3ce

I don't really care about keeping it or discarding. I might be leaning a bit towards ditching it. What's your take on this?

### DIFF
--- a/Benchmark/TestSQLiteStorePerformance.m
+++ b/Benchmark/TestSQLiteStorePerformance.m
@@ -231,8 +231,7 @@ static int itemChangedAtCommit(int i)
 
         NSString *expectedLabel = [self labelForCommit: rev child: i];
 
-        UKObjectsEqual(expectedLabel,
-                       [item valueForAttribute: @"name"]);
+        UKObjectsEqual(expectedLabel, item[@"name"]);
 
         // Step back one revision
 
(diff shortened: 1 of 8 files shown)


qmathe:
Forgot to mention I opened a related issue #48

ericwa:
Doesn't matter to me.. it's more concise, but it's a bit less clear because the symmetry between `-typeForAttribute`, `-valueForAttribute:` and `-setValue:forAttribute:ofType:` is obscured.

If `-valueForAttribute:` was called more commonly I would say apply the PR, but it's not used much so the verbosity isn't a huge issue, maybe we should just drop this PR.

qmathe:
Sounds reasonable. So let's close this PR and remove subscripting from COItem API in master.

Issue #65, opened 2016-09-27
Use const with local variables consistently


qmathe:
Currently we sometimes use const as a local variable qualifier when the type is a number like BOOL, Int, etc.

We should do a global search/replace and add/remove const where it makes sense.

ericwa:
IMO it makes sense in low level code. If there is any re-assigning / modifying of variables other than loop counters in the method, it's probably a good idea to use const.

qmathe:
ok. So I keep this issue opened until we take the time to review the code and apply const everywhere it makes sense.

PR #73, opened 2016-10-09
Add consts

jordyd:
I'm not done with this (so don't merge it just yet), I just wanted to open the PR for feedback purposes.

See also: #65

### DIFF
--- a/Extras/HistoryView/UIKit/COTrackViewController.m
+++ b/Extras/HistoryView/UIKit/COTrackViewController.m
@@ -107,7 +107,7 @@
 
 - (UITableViewCell *)makeCellForNode: (id <COTrackNode>)node
 {
-    static const NSString *const TrackCellIdentifier = @"TrackCell";
+    static NSString *const TrackCellIdentifier = @"TrackCell";
     UITableViewCell *cell = [self.tableView dequeueReusableCellWithIdentifier: TrackCellIdentifier];
 
     if (cell == nil)
--- a/Extras/HistoryView/UIKit/COTrackViewController.m
+++ b/Extras/HistoryView/UIKit/COTrackViewController.m
@@ -68,7 +68,7 @@
 - (id <COTrackNode>)nodeForRowAtIndexPath: (NSIndexPath *)indexPath
 {
     NSArray *nodes = self.track.nodes;
-    NSInteger reversedIndex = nodes.count - 1 - indexPath.row;
+    const NSInteger reversedIndex = nodes.count - 1 - indexPath.row;
 
     return nodes[reversedIndex];
 }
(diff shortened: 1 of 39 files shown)


Extras/HistoryView/UIKit/COTrackViewController.m
```diff
@@ -107,7 +107,7 @@ - (UIColor *)suggestedColorForNode: (id <COTrackNode>)node
 
 - (UITableViewCell *)makeCellForNode: (id <COTrackNode>)node
 {
-    static NSString *TrackCellIdentifier = @"TrackCell";
+    static const NSString *const TrackCellIdentifier = @"TrackCell";
```
jordyd: I made this const mainly because of that `static` there made me think of `FOUNDATION_EXPORT` lines with `static const NSString *const` values. But presumably in general Objective-C classes aren't to be made `const` (right?). So should this be `const` or no?

ericwa: I'm not sure, @qmathe do we want to use `static const NSString *const TrackCellIdentifier = @"...";`?

`const` is rarely used with ObjC objects in general, since there is no distinction in the type system between mutating and non-mutating ObjC method calls, like there is in C++.

jordyd: My two cents: I would remove the `const`s since (1) it would be consistent with other objects being non-`const`, and (2) making them `const` makes a promise that the compiler doesn't enforce. But for now I'll be adding `const`s in these kinds of situations just so that when you guys make a decision I can see where they all are in the commits.

jordyd:
I think that pretty much covers it. If you guys have any changes you want me to make let me know.

Extras/HistoryView/UIKit/COTrackViewController.m
qmathe: TrackCellIdentifier should be declared in a way that matches what we do for global string constants. So I would declare it:

`static NSString *const TrackCellIdentifier = @"TrackCell"`

iirc FOUNDATION_EXPORT is mapped to `extern` but not `extern const`.

qmathe:
I took a quick look at the changes, everything looks fine to me, except `static const NSString *const' instead of `static NSString *const`as discussed previously.

ericwa:
Yep - there are some `static const NSString *const` in `TestBinaryReadWrite` and `TestBinaryReadWritePerformance` that need to be updated, otherwise we can merge this, thanks!

ericwa:
@qmathe let's use the "Squash and Merge" option in github because we don't need 20 commits added to history for this change.

qmathe:
I Just closed the related issue #65.

Thanks @jordyd for your contribution! Thanks Eric for the merge!

PR #74, opened 2016-10-11
Remove support for 10.7

jordyd:
I also fixed some linker errors that appeared when switching to 10.8 deployment target. Note that my system is 10.12, so I can’t _fully_ test 10.8 compatibility.


jordyd:
It looks like the comment saying that macOS 10.7 doesn't allow you to add a pointer past the end of the array still applies at least in 10.9, given that's what Travis CI is running the tests with. It compiles and tests fine on my 10.12 machine though.

Undo/COUndoTrackStore.m
```diff
@@ -203,15 +203,6 @@ - (void)dealloc
     {
         [_db close];
     });
-
-#if !(TARGET_OS_IPHONE)
```
ericwa: Should change to `#if defined(GNUSTEP)`
and the start of the comment should be changed to `We are using deployment target 10.8, so ARC will manage libdispatch objects automatically. For GNUstep....`

Core/COCrossPersistentRootDeadRelationshipCache.h
```diff
@@ -56,10 +56,6 @@ NS_ASSUME_NONNULL_BEGIN
  * When a persistent root or branch is unloaded, any matching paths should be 
  * kept in the cache, in case it gets reloaded later (we cannot figure out cross 
  * persistent root incoming relationships from the reloading).
- *
- * For referring object graph contexts, when unloaded or finalized, the
- * deallocation will trigger their removal of their inner objects from the hash 
- * tables in the cache on 10.8 or iOS 6 or higher, but not on 10.7.
```
ericwa: I'd keep this comment, but change the end to 
>10.8 or iOS 6 or higher, which we require.

Store/COSQLiteStore.m
```diff
@@ -124,14 +124,6 @@ - (void)dealloc
         [db_ close];
         db_ = nil;
     });
-
-#if !(TARGET_OS_IPHONE)
```
ericwa: Should change to `#if defined(GNUSTEP)`
and the start of the comment should be changed to `We are using deployment target 10.8, so ARC will manage libdispatch objects automatically. For GNUstep....`

Debugging/COSQLiteStore+Graphviz.m
```diff
@@ -179,11 +179,8 @@ void COViewDOTGraphFile(NSString *dotFilePath)
             continue;
         }
 
-        // NOTE: Using NSTask rather than system() breaks 'po [objectGraphContext showGraph]' in LLDB on 10.7 
-        system([[NSString stringWithFormat: @"%@ -Tpdf '%@' -o '%@'",
-                                            executablePath,
-                                            dotFilePath,
-                                            pdfPath] UTF8String]);
+        [NSTask launchedTaskWithLaunchPath:executablePath
+                                 arguments:@[@"-Tpdf", dotFilePath, pdfPath]];
```
ericwa: Missing the `-o` param and `waitUntilExit`:
```
    NSTask *task = [NSTask launchedTaskWithLaunchPath:executablePath
                                                arguments:@[@"-Tpdf", dotFilePath, @"-o", pdfPath]];
    [task waitUntilExit];
```

qmathe:
@ericwa Well, my plan was to wait to get the test suite working again on GNUstep, before removing  any 10.7 support. Since I don't have the time to fix CO on GNUstep currently, it might be simpler to move forward and remove 10.7 support even if this causes us some extra work on GNUstep later.

I might have a bit more time next week to get CO working on GNUstep again. Not sure yet though.

My main worry was about the zeroing weak reference support on GNUstep. For example, if we remove `-[COCrossPersistentRootDeadRelationshipCache removePath:]. iirc this method is only needed if we don't have zeroing weak references.

Debugging/COSQLiteStore+Graphviz.m
```diff
@@ -179,11 +179,9 @@ void COViewDOTGraphFile(NSString *dotFilePath)
             continue;
         }
 
-        // NOTE: Using NSTask rather than system() breaks 'po [objectGraphContext showGraph]' in LLDB on 10.7 
-        system([[NSString stringWithFormat: @"%@ -Tpdf '%@' -o '%@'",
-                                            executablePath,
-                                            dotFilePath,
-                                            pdfPath] UTF8String]);
+        NSTask *task = [NSTask launchedTaskWithLaunchPath:executablePath
```
qmathe: A space is missing between launchedTaskWithLaunchPath: and executablePath. Same right after arguments:.

Tests/TestCommon.m
```diff
@@ -91,8 +91,6 @@ - (instancetype)init
 - (void)dealloc
 {
 #ifdef DELETE_STORE_AFTER_EACH_TEST_METHOD
```
qmathe: DELETE_STORE_AFTER_EACH_TEST_METHOD should now be defined by default or removed. @ericwa, what's your take on this?

Core/COCrossPersistentRootDeadRelationshipCache.m
```diff
@@ -32,14 +28,7 @@ - (void)addReferringObject: (COObject *)aReferrer
 
     if (referringObjects == nil)
     {
-        // FIXME: If we don't ditch 10.7 support, we need a reverse mapping
```
qmathe: This needs to be checked, I'm not 100% sure about what follows… I think we should either keep this comment somewhere or remove _referringObjectToPaths. If this FIXME is fully correct,  _referringObjectToPaths is probably only needed for `-removePath:` to work around missing support for zeroing weak references.

@ericwa, any comments on this point?

PR #75, opened 2021-10-29
Fix warnings/tests with recent macOS and add more generics annotations

qmathe:
- Warnings were caused by Apple deprecating more APIs starting from macOS 10.9.

- Test failure was caused by the default color space on macOS being changed from 'generic' (aka `NSCalibratedRGBColorSpace`)  to 'sRGB'. I'm not sure in which version that was changed, but the change I made is backward-compatible, so it shouldn't matter. Here is the failing test ouput:

> === [TestValueTransformerAttribute testRoundTrip] ===
> Etoile/Frameworks/CoreObject/Tests/Serialization/TestValueTransformerAttribute.m:101: warning: Failed, expected "sRGB IEC61966-2.1 colorspace 1 0 0 1", got "NSCalibratedRGBColorSpace 0.984314 0 0.0235294 1"

- The new generics annotations allow to call more CoreObject APIs from Swift without casting.

- For sanboxed apps on macOS, posting distributed notifications containing a user info dictionary (and also registering for distributed notifications without targeting a specific object) causes sandboxing violations:

> Toucan *** attempt to post distributed notification 'COStorePersistentRootsDidChangeNotification' thwarted by sandboxing.
> 
> Date/Time:     Fri Oct 29 15:10:26 2021
> OS Version:    21A559
> Application:   Toucan
> 
> Backtrace:
> 0   CoreFoundation                      0x00007ff80ddb75b8 __CFGenerateReport + 175
> 1   CoreFoundation                      0x00007ff80dc4e9f8 _CFXNotificationPost + 1359
> 2   Foundation                          0x00007ff80eb65f5b -[NSDistributedNotificationCenter postNotificationName:object:userInfo:options:] + 84

My solution to this is to use a custom distribution center that does nothing when sandboxing is enabled. The previous solution introduced for iOS doesn't work, because NSDistributedNotificationCenter is a public API on macOS unlike iOS, so the compiler ignores the compatibility alias.


ericwa (approved):
I haven't run anything, but changes look fine!

qmathe:
Thanks for the review, Eric. 

I took a look at PR #74 yesterday. Before merging, I'd like to take most changes from PR #74 and apply them on top of this PR, since both PRs are related. I could then close PR #74 without applying it. Any objections?

ericwa:
No objections!

qmathe:
Oops, I didn't run the tests with -DSANDBOXED=1 last time, so `CODistributedNotificationCenter` implementation was wrong and some tests were broken. I fixed it. The code now covers for the case where we sync two editing contexts backed by two distinct store instances pointing to the same underlying SQLite db on disk.

I also applied changes from PR #74 on top.

In `COCrossPersistentRootDeadRelationshipCache`, we have this comment:

```
        // FIXME: If we don't ditch 10.7 support, we need a reverse mapping
        // from each referringObject to a path set, that can be used to remove
        // the referring objects when their object graph context is discarded.
```


For me, it means `_referringObjectToPaths` can be removed if we support weak-zeroing references as it is the case on 10.8 and higher (not sure about GNUstep though). I tried to do it, but I get a test failure and some exceptions, so not sure what's going on here. I'll keep this for later, since I don't want to spend too much time on it for now.

qmathe:
@ericwa Any comments on my latest changes before I merge?

ericwa:
No special comments, looks fine.

PR #76, opened 2023-02-26
Upgrade to project to Xcode 14.2 and fix warnings

qmathe:
Hi Eric,

I fixed warnings with the last Xcode version and updated the project to use the latest supported deployment targets.

I turned off a new warning about 'missing explicit self reference inside block', because it was causing hundreds of warnings. To fix it would require to add `self->property` in block-based tests, but I find it hurts readability, since we're using blocks everywhere in the tests.

There is one remaining warning: _TestMetamodelCornerCases.m:57:44 Undeclared selector 'setAggregate:'_. Although the documentation above `-testNoAggregate` does make sense to me, I don't get what the test is testing, since there is no `-setAggregate:` method in the metamodel.

### DIFF
--- a/Core/COObject+Private.h
+++ b/Core/COObject+Private.h
@@ -19,7 +19,7 @@ BOOL IsSetter(const char *selname, size_t sellen);
 BOOL isSerializablePrimitiveValue(id value);
 BOOL isSerializableScalarValue(id value);
 
-ETEntityDescription *entityDescriptionForObjectInRepository();
+ETEntityDescription *entityDescriptionForObjectInRepository(id anObject, ETModelDescriptionRepository *repo);
 
 @interface COObject ()
 
--- a/Core/COObject.m
+++ b/Core/COObject.m
@@ -777,7 +777,7 @@ static NSNull *cachedNSNull = nil;
 {
     ETPropertyDescription *propertyDesc = [_entityDescription propertyDescriptionForName: key];
     ETPropertyDescription *opposite = propertyDesc.opposite;
-    ETValidationResult *oppositeResult = nil;
+    __unused ETValidationResult *oppositeResult = nil;
 
     if (opposite != nil)
     {
--- a/CoreObject.xcodeproj/project.pbxproj
+++ b/CoreObject.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 50;
+	objectVersion = 54;
 	objects = {
 
 /* Begin PBXBuildFile section */
@@ -2959,7 +2959,7 @@
 			isa = PBXProject;
 			attributes = {
 				CLASSPREFIX = "";
-				LastUpgradeCheck = 0700;
+				LastUpgradeCheck = 1420;
 				ORGANIZATIONNAME = "Étoilé";
 				TargetAttributes = {
 					60E08C6119792BEA00D1B7AD = {
@@ -2972,13 +2972,9 @@
 			};
 			buildConfigurationList = C01FCF4E08A954540054247B /* Build configuration list for PBXProject "CoreObject" */;
 			compatibilityVersion = "Xcode 9.3";
-			developmentRegion = English;
+			developmentRegion = en;
 			hasScannedForEncodings = 1;
 			knownRegions = (
-				English,
-				Japanese,
-				French,
-				German,
 				Base,
 				fr,
 				en,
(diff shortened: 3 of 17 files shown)


qmathe:
While running the tests, I also found that output prints this error: 

_BUG IN CLIENT OF libsqlite3.dylib: database integrity compromised by API violation: vnode unlinked while in use: /private/var/folders/h6/j74mwvkx2v1g7_dhhrh1y5_80000gn/T/TestUndoTrackStore.sqlite/undo.sqlite-wal_

It looks the database is not closed correctly, but `COUndoTrackStore` does close it in -dealloc, so this should work fine. I guess a reference to a COUndoStackStore is not released somewhere. I haven't looked into it very deeply, I'll do it later. If you have an idea about what's going on, let me know.

ericwa:
Hi Quentin,

> There is one remaining warning: TestMetamodelCornerCases.m:57:44 Undeclared selector 'setAggregate:'. Although the documentation above -testNoAggregate does make sense to me, I don't get what the test is testing, since there is no -setAggregate: method in the metamodel.

Yeah, I think it's safe to delete that assertion since there is no `setAggregate:` in the codebase.

> BUG IN CLIENT OF libsqlite3.dylib: database integrity compromised by API violation: vnode unlinked while in use: /private/var/folders/h6/j74mwvkx2v1g7_dhhrh1y5_80000gn/T/TestUndoTrackStore.sqlite/undo.sqlite-wal

As a guess, when https://github.com/etoile/CoreObject/blob/master/Tests/TestCommon.m#L98 runs to clean up test data, maybe we still have the database open in some background worker thread?

ericwa (approved):

qmathe:
> Yeah, I think it's safe to delete that assertion since there is no setAggregate: in the codebase.

ok, I removed it and moved up the comment at the top. I kept the comment because it gives another perspective to what part of EMOF/UML we implemented compared to `ETModelElementDescription` documentation.

As a side note, I have been thinking about this part, because COCopier in its current implementation is broken when non-composite cycles appear in it. When encountering a cycle, it copies the target object but not the relationships of the target object. If you copy the relationships of the target object, it's hard to know when to stop and you can end up copying the entire object graph. 
My current solution is to disable copying non-composite relationships. I'm unable to come up with an automatic approach to handle object graph copying, because it usually depends on how you plan to use and insert the copied objects. I think the only way to handle this is to declare copying-related annotations in the metamodel for non-composite relationships. What do you think?

> As a guess, when https://github.com/etoile/CoreObject/blob/master/Tests/TestCommon.m#L98 runs to clean up test data, maybe we still have the database open in some background worker thread?

Thanks for the suggestion. I'll check this and open another PR once I have tracked it down.

qmathe:
Hey Eric,

I just merged this PR (working on Toukan again after a six months break). 

What's your take about the COCopier issue described in my previous comment? I'd be curious to hear your opinion, since I  cannot figure out a clear solution.

qmathe:
Your point about copying the entire object graph being an an architecture issue the user has to deal with makes sense. If Unreal team thought this was an acceptable trade-off, its proves your point and I guess it implies it's going hard to do better. 
We could come up with more complex rules to cope with it, but in the end complexity will outweigh benefits, since these situation where the entire object graph can be copied are edge cases.

> For how this applies to CoreObject: I'd say if the copy destination is in the same object graph, you don't want any non-composite references to be copied, but rather just reuse those existing objects.

For Toukan, I pushed back support for shared assets to after the first version, which means almost every single relationship is a composite in a document, so copying is easy.

The only exception being when a layer sampler is referencing another layer somewhere in the document tree. In this case, I must do a cycle detection when creating the sampler to prevent an infinite rendering loop where the sampler samples itself through a parent or ancestor layer. 
For copying a layer sampler, I'm not sure what's the best approach. I have a user experience issue to solve, when the user deletes the source layer or copies the sampler to another document. In both cases, I have to decide what I do with the source layer. I could either:

1. reinsert as a top hidden layer in the document root
2. keep referencing it in the sampler even it's missing in the document tree

The latter case prevents the user from editing it, because supporting editing a layer not present in the document tree would substantially increase the app complexity (it's something I prefer to avoid for the first version). In the long run, I could have an asset library (images, layers, colors etc.) inside each document, then I could turn the sampled layer into an asset and support editing layers stored as assets.

I think the best option is the first one. I could even check the inverse relationship cache to find whether any samplers are referencing the layer to delete. If it's the case, I can ask the user whether he wants to discard it completely or move it to the top as a hidden layer, so the layer sampler doesn't break. 
For copying, I can copy the source layer even though it's not a composite relationship, but ignore it at pasting time when already present the destination object graph.

What do you think? 

> The option of stopping after copying "1 layer deep" of non-composite references doesn't really make sense to me; it seems like you'd just end up with broken objects/missing data.

Agreed.

> I don't know how exactly this would integrate with cross-persistent-root references, maybe that would be a user facing option to create cross-persistent-root references instead of copying non-composite referenced objects?

For cross-persistent references, I would simply keep them as cross-persistent refs instead of copying the referenced objects.

PR #77, opened 2023-10-10
Nullability and lightweight generics annotations for Store, Undo and Localization

qmathe:
To improve Swift compatibility, I added annotations to headers in Store, Undo and Localization directories. 

The nullability annotations are based on returned value and argument assertion checks in the implementation files.

I also renamed _replcatedBranch_ to _replicatedBranch_ for Synchronization related code. Since existing mirrored branches have the wrong name, I expect this change to cause previously mirrored branch to be lost for CoreObject demo apps. I don't know whether it's an issue or not. I can remove the change or push it back to later if needed.


ericwa (approved):

qmathe:
Thanks for the quick review!

PR #78, opened 2023-10-13
Make COUndoTrack usable without COEditingContext

qmathe:
Hey Eric,

I'm resuming my work on Toukan and CoreObject today.

To give you some background about this PR… In Toukan, I have a high-level API written in Swift, that replaces COEditingContext, COPersistentRoot, COBranch, COObjectGraphContext and COObject. It's much more limited in term of features and scope. I more or less reduced the API to COEditingContext and CObjectGraphContext.

For now, I don't have support for branches or cross-persistent references, which I might reintroduce later to not. I also replaced the metamodel and relationship cache with property annotations. For example, declaring a to-many  composite relationship for a property 'children' look like this:

```swift
    @RelationshipToMany<Node, Node>(inverse: \Node.parent, isParent: true, sendsChange: false)
    var children: [Node]
```

---

In the current CoreObject version, COUndoTrack is partially tied to the Core API. With this PR, I change COUndoTrack and related classes (COCommand class hierarchy and COUndoTrackStore) to become completely independent of the Core classes (COEditingContext, COObjectGraphContext etc.), so it can be used with other high-level API such as the one I wrote in Swift for Toukan.

Here are the key changes I made in this PR:
- Made `CORevisionCache` independent of `COEditingContext`
  - when writing a replacement for the Core API, it's now possible to reuse it
  - the revision cache now depends on `COSQLiteStore`
- Make `COCommand` independent of `COEditingContext`
  - a `COUndoTrackContext` protocol has been introduced and  can be implemented to double-dispatch operations on the undo track to the right command through the context
  - `COUndoTrackContext` is declared in _COUndoTrack.h_
  - see `-applyCommand:` and `-applyCommand:toStoreTransaction:withMetadata:`
- Made `COLeastCommonAncestor` algorithms independent of `COEditingContext`
  - these methods have been turned into standalone functions that takes a parent revision provider
  - when writing a replacement for the Core API, it's now possible to reuse them
- New `-[CODiffManager diffItemWithItemGraph:algorithmClasses:sourceIdentifier:]` to support making diffs without a model description repository
  - in Toukan, model classes own their metamodel (eliminating the need to construct a standalone metamodel)
    - attributes/relationships are described with annotations attached to properties
    - annotations can be accessed at run-time through model instances

### DIFF
--- a/Benchmark/TestHistoryNavigationPerformance.m
+++ b/Benchmark/TestHistoryNavigationPerformance.m
@@ -68,7 +68,7 @@
 - (void)testGoToOldestAndNewestNodesInHistory
 {
     COUndoTrack *track = [COUndoTrack trackForName: @"TestHistoryNavigationPerformance"
-                                withEditingContext: ctx];
+                                withContext: ctx];
 
     NSDate *startDate = [NSDate date];
     NSArray *proots = [self commitPersistentRootsWithUndoTrack: track
@@ -130,7 +130,7 @@
 - (void)testGoToFirstCommitNodeToReloadManyDeletedPersistentRoots
 {
     COUndoTrack *track = [COUndoTrack trackForName: @"TestHistoryNavigationPerformance"
-                                withEditingContext: ctx];
+                                withContext: ctx];
 
     NSDate *startDate = [NSDate date];
     NSArray *proots = [self commitPersistentRootsWithUndoTrack: track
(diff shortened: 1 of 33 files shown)


ericwa (approved):

qmathe:
Thanks for the quick review, Eric!

PR #81, opened 2024-02-15
Shorter internal key prefixes in COItem

qmathe:
Hey Eric,

Existing default key prefixes in COItem are quite long. When working with big object/item graphs in Toukan, I found this makes debugging output more verbose and harder to go through quickly. So I'd like to remove the reverse DNS key prefix and simply use an underscore as prefix. The underscore as prefix would be reserved for CoreObject.

For the short term, I changed the deserialization logic to check whether an item is encoded with the old prefixes or the new ones. When old prefixes are used, I manually replace the old keys with new keys in both `COItem.values` and `COItem.types` dictionaries. This replacing is probably going to hurt performance when loading big item graphs, but I'm ok with it in the short term. My plan is to keep in the coming months, then remove it once:
- everything works well
- documents have been all snapshotted with these new prefixes
- past document histories have been compacted (eliminating items in store using the old prefixes).

Once I remove it, it would break compatibility with existing CoreObject stores you might have still around. Is it an issue for you?

---

The new current internal key prefixes common to both JSON and Binary formats are:
- _entityName
- _packageName
- _packageVersion

Two more internal keys exists only for JSON format:
- _uuid
- _json-format

As you can see, there is a slight mismatch in term of naming styles:
- dash -> _json-format
- camel case -> _entityName

I could use dash everywhere for internal keys:
- _entity-name
- _package-name
- _package-version
- _uuid
- _json-format

Not sure whether it's a good idea, since we use camel case in the API and when serializing ObjC/Swift objects. If we were serializing model objects in other languages or porting CO to other languages, the property naming could be in another style, so standardizing on ObjC/Swift naming style might not be important.

The JSON format also encodes the multivalued property types using dash. For example: _string-array_

So I see 5 possible choices:
1. leave it as it is in the PR currently
2. name every internal keys with dash
3. rename *_json-format* to *_jsonFormat*, but leave other multivalue type keys as they are
4. name every internal keys with camel case (including multivalued type keys in JSON format)

Given that choice 3 creates incoherencies inside the JSON format and 4 is way too much work and troubles, I'd go for 1 or 2. What do you prefer?

### DIFF
--- a/StorageDataModel/COItem+JSON.m
+++ b/StorageDataModel/COItem+JSON.m
@@ -16,8 +16,10 @@
 // Semi-hack: we store the UUID alongside the real object properties.
 // This property name is reserved for JSON serialization and cannot be used
 // as an actual property name.
-NSString *const kCOJSONObjectUUIDProperty = @"org.etoile-project.coreobject.uuid";
-NSString *const kCOJSONFormatProperty = @"org.etoile-project.coreobject.json-format";
+NSString *const kCOJSONObjectUUIDProperty = @"_uuid";
+NSString *const kCOJSONFormatProperty = @"_json-format";
+NSString *const kCOJSONDeprecatedObjectUUIDProperty = @"org.etoile-project.coreobject.uuid";
+NSString *const kCOJSONDeprecatedFormatProperty = @"org.etoile-project.coreobject.json-format";
 NSString *const kCOJSONFormat1_0 = @"1.0";
 // COType -> string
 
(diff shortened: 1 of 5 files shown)


ericwa:
> Once I remove it, it would break compatibility with existing CoreObject stores you might have still around. Is it an issue for you?

This is fine, I don't have any existing ones.

For naming style, between option 1 and 2, I would go with 2, since it gets rid of the `_json-format` and `_entityName` clashing naming style.

qmathe:
Thanks for the review! For the naming style, I went for option 2 as you suggested.

PR #83, opened 2024-02-28
New schema migration to rewrite revisions in store

qmathe:
Hey Eric,

Here is a first attempt at rewriting revisions for schema migration. 

I decided to make schema migration an all-or-nothing operation for the entire store. I did this to keep the API simpler and eliminates issues when moving items between persistent roots with mismatched schema versions (e.g. copy/paste or cross-persistent root references).

I added a notion of current schema version to _storeMetadata_ table. While working on this, I added simple migration model for the store schema, to ensure my existing stores get automatically updated. 

The store can either enforce a schema version (refusing writing a revision with a different schema version) or not (accepting writing a revision with a schema version higher than the current one), see `COSQLiteStore.enforcesSchemaVersion`. The goal is to support migrating item graphs  at loading time as an alternative to rewriting the store. For now, `COSQLiteStore.enforcesSchemaVersion` is NO by default. Do you think it should be YES by default?

To make debugging easier, I added a column _schema_version_ to backing stores rather than storing it in the revision metadata. As a result, some APIs now expose an explicit schema version (e.g. CORevision, CORevisionInfo, COStoreTransaction etc.).

The current schema version when committing on a branch is taken from `ETModelDescriptionRepository.version` (this is a new property I added). In most cases, I expect `ETPackageDescription.version` and `ETModelDescriptionRepository.version` to be set to the same value. If you have multiple packages, versioning them separately remains possible but you must bump the repository schema version every time a package version changes. I think it's a good idea to keep `COItem.version` in addition to `OSQLiteStore.schemaVersion`, since COItem(s) can easily be checked to ensure their version is what you expect.

As a side note, the store schema version applies to the item graphs / revisions, while the store version (from _storeMetadata_ table) applies to the store table/column schema. Not sure my terminology is the best here, could be better to swap their usage or figure out something else.

Since I'm rewriting all revisions in a single pass per backing store, I didn't bother with rewriting child commits as you were proposing, because it would slow down the rewrite (by rewriting the child commit twice). I pondered about adding a distinct `-rewriteRevid:withItemGraph:` to the backing store API, but I doubt it's useful without a concrete use case. When I consider "deleting passwords" from history as a use case, I suppose it would require to rewrite all revisions.

`COSchemaMigrationDriver` could be reused inside the handler of `-migrationRevisionsToVersion:withHandler:`, but I'm in favor of discarding it (except the documentation that covers some interesting point). Most of the complexity from `COSchemaMigrationDriver` comes from the possibility to migrate entities that are subtypes of other entities in another package. I'm now convinced this should be avoided, because migrating such entities is error-prone. 
If you really want to reuse entities across apps, each app can expose a framework containing the migration logic for its own package and an adhoc solution similar to a `COSchemaMigrationDriver` could be created for this precise use case.

I'm planning to write some more tests, but I'd be interested to have your feedback on what I have now.


Store/COSQLiteStorePersistentRootBackingStore.m
```diff
@@ -643,6 +659,106 @@ - (BOOL)writeItemGraph: (COItemGraph *)anItemTree
...
+                version: (int64_t)newVersion
+     migratedItemGraphs: (NSDictionary *)migratedItemGraphs
+    lastBytesInDeltaRun: (int64_t)lastBytesInDeltaRun
+{
+    FMResultSet *rs = [db_ executeQuery:
+        [NSString stringWithFormat: @"SELECT parent, deltabase FROM %@ WHERE revid = ?", [self tableName]],
+        @(revid)];
+    int64_t bytesInDeltaRun = -1;
+
+    while ([rs next])
```
ericwa, line 673: Just a style nitpick, this can only return 0 or 1 rows (0 means the caller provided an invalid `revid`), the `while` loop suggests it can return several rows at first glance. Maybe just an `if` would be clearer (and report an error if 0 rows were returned?)

Store/COSQLiteStorePersistentRootBackingStore.m
```diff
@@ -643,6 +659,106 @@ - (BOOL)writeItemGraph: (COItemGraph *)anItemTree
...
+- (BOOL)migrateRevisionsToVersion: (int64_t)newVersion withHandler: (COMigrationHandler)handler
+{
+    NSIndexSet *revids = [self revidsUsedRange];
+    NSMutableDictionary *migratedItemGraphs = [NSMutableDictionary dictionary];
+    int64_t __block lastBytesInDeltaRun = 0;
+
+    [self beginTransaction];
+
+    [revids enumerateIndexesUsingBlock: ^(NSUInteger revid, BOOL * _Nonnull stop) {
+        COItemGraph *oldItemGraph = [self itemGraphForRevid: revid];
```
ericwa, line 738: I think there's a bit of a problem here because `rewriteRevid:` doesn't implement the full rewrite logic that we were discussing (where child delta commits are fixed so they can return properly return the "old" commit contents).

Say you are migrating all 3 commits in the store, A (full snapshot), B (delta), C (delta).

This block runs for A, B, C, in that order.

Everything works fine for revision A, and it gets rewritten.

But when the block runs for B, this `-itemGraphForRevid: B` call is going to return a mix of the already-migrated items from A  plus some of the unmigrated items from B. Won't that be a problem for the `handler()` function?


I think if you want to go with the present approach here you'll need to use `-partialItemGraphFromRevid:toRevid:` to get *just* the items in the delta commit, that way you're guaranteed not to have a mix of item schema versions unlike with `-itemGraphForRevid:`.

With the approach in this PR it also means that `handler()` may be called with an incomplete item graph, so I think you can only support the "simple" type of migrations that are 1 item in -> the same 1 item out.

ericwa:
> As a side note, the store schema version applies to the item graphs / revisions, while the store version (from storeMetadata table) applies to the store table/column schema. Not sure my terminology is the best here, could be better to swap their usage or figure out something else.

Yes, it's a bit confusing, it's hard to know whether "schema" / "version" refers to the item schema, or the COSQLiteStore file format / SQL table schema.

Could cal them "item schema" or "user schema" vs "CoreObject store file format version" or something like that.

ericwa:
> COSchemaMigrationDriver could be reused inside the handler of -migrationRevisionsToVersion:withHandler:, but I'm in favor of discarding it (except the documentation that covers some interesting point)

That's fine with me, I never really grasped the old migration system to be honest.

Store/COSQLiteStorePersistentRootBackingStore.m
qmathe: Yes, this while statement doesn't make it clear there is a single result. I'm going to change it. If an invalid revid is provided, I plan to log an error and return -1. 

For now, we don't have error propagation in `COSQLiteStore` and `COSQLitePersistentRootBackingStore`, that's why I'd like to stick this style (log error and return -1 or NO). In the long run, it could be better to refactor the whole implementation to propagate detailed errors up the public API that is called.

qmathe:
> Yes, it's a bit confusing, it's hard to know whether "schema" / "version" refers to the item schema, or the COSQLiteStore file format / SQL table schema.
> 
> Could cal them "item schema" or "user schema" vs "CoreObject store file format version" or something like that.

Taking about _format_ for the internal store structure sounds good, then we can keep using _schema_ to talk about the user data structure. I don't think we need to be more precise with _item schema_ or _user schema_, since the store format is not even something we need to document publicly. I can add a comment where the _storeMetadata_ table is created to clarify the differences between both.

Store/COSQLiteStorePersistentRootBackingStore.m
qmathe: > But when the block runs for B, this -itemGraphForRevid: B call is going to return a mix of the already-migrated items from A plus some of the unmigrated items from B. Won't that be a problem for the handler() function?

Thanks for pointing this out! I wrongly assumed that inside the transaction, queries would read the DB in the state prior to the transaction, but I checked SQLite documentation and that's not true: _Within a single database connection X, a SELECT statement always sees all changes to the database that are completed prior to the start of the SELECT statement, whether committed or uncommitted._

It's strange my migration test doesn't fail due to this, I'll need to take a better look.

To fix this, I need to either:

1. use the rewrite logic you suggested initially
2. rewrite revisions backwards (from most recent revision to the oldest one)

Option 1 is probably a bit slower because rewriting affects two revisions and each revision is rewritten twice. Option 2 is more complex to implement, because you need to pick a branch and a head revision to start with, then you have to check at every revision whether there is a child branch that needs to be rewritten first, before rewriting the previous revision on the branch you're processing.

I guess that's why you suggested option 1 initially.

ericwa: Yeah, I suggested option 1 because it keeps the store in a consistent state. You’re right it will rewrite each revision twice though so it is potentially a lot slower.

I think you can implement option 2 with a tiny change to your current code, just iterate the revids backwards (highest to lowest).

qmathe:
I changed `-[COPersistentRootBackingStore migrateToVersion:withHandler:]` to rewrite the revisions backwards, in order to ensure retrieving full item graphs for delta revisions remain possible during the migration (without getting mixed item versions). 

I also wrote some more tests to cover migrating multiple persistent roots with a single branch and multiple branches for a single persistent root. Do you think I should write a more complex test such multiple branches in multiple persistent roots?

Beside this, I fixed the benchmark which was broken due my COSQLiteStore API changes and wrote a new benchmark for revision rewriting. On my machine, revision rewriting ends up being quite slow (roughly 1000 revisions per second with an item graph containing 50 items and touching 1 or 2 items per revision). For a big store with 500 documents containing 1000 revisions each, it's going to take around 8mn.
When I run the benchmark on my MacBook Air M1, I get the following result:

=== [TestRevisionRewritingPerformance testFixedItemCount] ===
Time to commit 50 new persistent roots: 0,3140s
Time to make 99 commits: 1,9039s
Time to rewrite 5 000 store revisions: 5,1078s

=== [TestRevisionRewritingPerformance testGrowingItemCount] ===
Time to commit 50 new persistent roots: 0,0224s
Time to make 99 commits: 2,7854s
Time to rewrite 5 000 store revisions: 5,1270s

=== [TestRevisionRewritingPerformance testSingleItem] ===
Time to commit 50 new persistent roots: 0,0235s
Time to make 99 commits: 1,4526s
Time to rewrite 5 000 store revisions: 0,5050s

Could you check the changes I made to COSynchronizer and COCommand related classes make sense? I added a schema version to COSynchronizerRevision and I now assert schema version is the same before rebasing/merging revisions.

qmathe:
While testing one of the new benchmark I wrote, I observed that when `Person.students` contains more than around 250 elements, there is a crash in the binary writer when turning the item graph into data. I attached a screenshot that shows where the exception arises (switching on `type' due to an unexpected null terminator). I don't understand what is going on and I'm not familiar with the binary reader/writer code. Any ideas?

Capture d’écran 2024-03-04 à 13 57 21

ericwa:
I think I see the bug, `writeSetContents` is saving pointers into `co_buffer_t *temp` into the `tokenPointers` array, but these pointers become invalid when the `temp` buffer resizes (when it exceeds the initial allocation `#define CO_BUFFER_INITIAL_LENGTH 4096`).

The overall idea of `writeSetContents` is it writes the contents of the set being serialized into `temp` in a random order, then it sorts the pointers to the start of each "token" (the serialized form of each set element), then copies the serialized elements from `temp` in the final order into the output `dest`. (The reason it's calling `co_reader_length_of_token` is, it's not storing the lengths of the serialized tokens, it's recomputing them as needed).

The fix is `tokenPointers` should be changed from an array of pointers to an array of integer offsets from the start of `temp`.
The `comparePointersToBinaryWriterTokens` comparison function will need to be updated, and instead of `qsort` we'll need `qsort_s` or `std::sort` so we can pass in the start of the bufffer as the context parameter.

qmathe:
> The fix is `tokenPointers` should be changed from an array of pointers to an array of integer offsets from the start of `temp`. The `comparePointersToBinaryWriterTokens` comparison function will need to be updated, and instead of `qsort` we'll need `qsort_s` or `std::sort` so we can pass in the start of the bufffer as the context parameter.

ok, thanks for the explanation. I understand better how this works. I'll try to implement the fix you outlined.

ericwa:
> Could you check the changes I made to COSynchronizer and COCommand related classes make sense? I added a schema version to COSynchronizerRevision and I now assert schema version is the same before rebasing/merging revisions.

Seems fine.. in general (iirc you mentioned the same thing) I think anything with diff/merge (undo tracks, synchronizer) will break if there are different schemas involved. So the assertions make sense.

qmathe:
I removed the old schema migration code this morning, documented `-migrationRevisions:withHandler:` and I just fixed the crash in `writeSetsContents()`. Should be ready for merging now.

PR #84, opened 2024-03-07
Add options to control COCopier behavior and fix copying non-composite references

qmathe:
The 3 key changes are:
- support copying non-composite references to arbitrary depths (possibly copying the full source item graph)
- new copying behavior options
  - reuse existing source item UUIDs rather than creating new ones
    - can be used to easily clone some part of an item graph (e.g. a composite item tree) without changing its identity
  - choose between copy or alias for non-composite references that exist in the destination
  - choose between copy or ignore for non-composite references that are missing in the destination

### DIFF
--- a/Core/COCopier.h
+++ b/Core/COCopier.h
@@ -16,12 +16,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 typedef NS_OPTIONS(NSInteger, COCopierOptions) {
     /**
-     ** Whether copies have different UUIDs in destination than original items in source.
+     * Whether copies keep using the same UUIDs in destination than original items in source.
      *
-     * This is useful to duplicate items when source and destination are the same. You can also use 
-     * it to implement copy/paste semantics whether source and destination are the same or not.
+     * You should omit this option to implement cut/copy/paste semantics, whether source and 
+     * destination are the same or not.
      */
-    COCopierUsesNewUUIDs = 2,
+    COCopierReusesSourceUUIDs = 2,
     /**
      * Whether items reachable through non composite references are copied, when they dont'
      * exist in the destination.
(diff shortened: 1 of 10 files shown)


Core/COCopier.h
```diff
@@ -11,6 +11,39 @@
...
+     * exist in the destination.
+     *
+     * Be careful with this option, the entire source item graph can be copied into the destination.
+     *
+     * Can be used alone and in combination with
+     * COCopierCopiesNonCompositeReferencesExistingInDestination.
+     */
+    COCopierCopiesNonCompositeReferencesMissingInDestination = 4,
+    /**
+     * Whether items reachable through non composite references are copied, when they exist in the
```
ericwa, line 36: Copying items that exists in the destination means updating the item to match the state in the source item graph I guess?

ericwa (approved):
I just skimmed the implementation but this sounds good!

Core/COCopier.h
qmathe: It depends whether you pass COCopierReusesSourceUUID. 

If you pass COCopierReusesSourceUUID, then the item in the destination is updated to match the item in the source graph.

If you don't pass  COCopierReusesSourceUUID, then the corresponding item in the destination is left untouched, because the item in the source graph is copied into the destination with a different UUID.

I'll add a note about this point to make it clearer.

qmathe:
I updated the API documentation based on your feedback and I also documented how the new options interacts with kCOIsSharedItemProperty.

In addition to this, I rewrote the copy algorithm to collect item UUIDs with a single traversal. The previous implementation was using two distinct recursive functions to collect composite references and non-composite references. The resulting code is shorter and should be faster.

